PR #124, opened 2023-04-04
Update Nuget.exe default version

zhenghao104:
Update Nuget.exe default version and only use supported msbuild version


common-npm-packages/packaging-common/nuget/NuGetToolGetter.ts
```diff
@@ -184,16 +184,18 @@ export async function cacheBundledNuGet(
...
-            taskLib.debug('Snapping to v5.4.0');
-            cachedVersionToUse = '5.4.0';
-            nugetPathSuffix = 'NuGet/5.4.0/';
+        // Default to 6.5.0 if we're using MSBuild 17.0.0 or higher
+        // Default to 5.9.3 if we're using MSBuild 16.11.0 or higher, older MSBuild are not supported
+        // Default to 4.9.4 if we're using MSBuild older than 16.11.0
+        if (msbuildSemVer && semver.gte(msbuildSemVer, '17.0.0')) {
+            taskLib.debug('Snapping to v6.5.0');
+            cachedVersionToUse = '6.5.0';
+            nugetPathSuffix = 'NuGet/6.5.0/';
```
phil-hodgson, line 194: According to https://dist.nuget.org/tools.json, 6.5.0 is "Released" but not yet "ReleasedAndBlessed". Check with NuGet folks on where they are with verifying/blessing 6.5.0. If it'll be a while, then we can switch to 6.4.0 in the short term.

common-npm-packages/packaging-common/nuget/NuGetToolGetter.ts
```diff
@@ -184,16 +184,18 @@ export async function cacheBundledNuGet(
...
+        } else if (msbuildSemVer && semver.gte(msbuildSemVer, '16.11.0')) {
+            taskLib.debug('Snapping to v5.9.3');
+            cachedVersionToUse = '5.9.3';
```
phil-hodgson, line 197: Do we need to update NuGetCommand to bundle 5.9.3 (and 6.x.0)? so that they're always available with the task?

common-npm-packages/packaging-common/nuget/NuGetToolGetter.ts
phil-hodgson, line 1: Is there any test reaction work?

common-npm-packages/packaging-common/nuget/NuGetToolGetter.ts
zhenghao104: If we run nuget tool installer without specifying a version, it will use 6.5.0

common-npm-packages/packaging-common/nuget/NuGetToolGetter.ts
zhenghao104: Yes, we need to push the 5.9.3 and the 6.x.x version we decide to vstsblobagent.

common-npm-packages/packaging-common/nuget/NuGetToolGetter.ts
zhenghao104: It may be difficult to explicitly test different msbuild/nuget version combination with just L0 tests

phil-hodgson: There's testing options for this repo under `./Tests` (e.g. `./Tests/L0.ts`)

zhenghao104: We have L0 tests here for nuget
https://github.com/microsoft/azure-pipelines-tasks-common-packages/blob/main/common-npm-packages/packaging-common/Tests/nugetL0.ts
But I m not sure if it's possible to test this specific change since it requires a specific version of msbuild and nuget.exe being available at runtime to test its compatibility, this sounds more like a L2 tests

phil-hodgson: It should be possible to mock the file system checks. The task library includes the ability to setup mock responses (see [here](https://github.com/microsoft/azure-pipelines-task-lib/blob/f7e17959900884f773e65b0e3d3d81ba69b6371f/node/mock-task.ts#L12))

Here's an example of it being used: https://github.com/microsoft/azure-pipelines-tasks-common-packages/blob/36425c4984bfd90d3d2e777c437413acf3f18cc6/common-npm-packages/securefiles-common/Tests/L0.ts#L80

zhenghao104: Added two tests. The way that we check for msbuild version makes it difficult to be mocked since it tries to set a variable that's out of scope. A few learnings here, we should add L0 tests for all methods that we exported and refactor the code if needed to be mockable. We should consider using `jest` instead of `mockery` as `jest` is a very popular library with great documentation and support.

common-npm-packages/packaging-common/nuget/NuGetToolGetter.ts
```diff
@@ -161,16 +163,13 @@ export async function getMSBuildVersionString(): Promise<string> {
...
-        getVersionTool.on('stdout', (data: string) => {
-            if (data) {
-                version = data.toString().trim();
-                taskLib.debug('Found msbuild version: ' + version);
-            }
-        });
-        await getVersionTool.exec();
-        taskLib.debug('Finished running msbuild /version /nologo');
+        try {
+            const msbuildVersion: VersionInfo = await peParser.getFileVersionInfoAsync(path);
```
zhenghao104, line 167: This will returns sufficient information for semver comparison.

jmyersmsft (approved):

common-npm-packages/packaging-common/nuget/NuGetToolGetter.ts
```diff
@@ -161,16 +163,13 @@ export async function getMSBuildVersionString(): Promise<string> {
...
+            version = msbuildVersion.productVersion.toString();
+        }
+        catch (err) {
+            taskLib.debug('Unable to find msbuild version');
+        }
     }
     return version;
```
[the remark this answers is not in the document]
zhenghao104: Added

embetten (approved):

phil-hodgson (approved):

PR #145, opened 2023-05-05
Added remaining Unit Tests

sergey-koryshev:
Implemented remaining Unit Tests


common-npm-packages/codecoverage-tools/Tests/coberturaantccenablerTests.ts
```diff
@@ -0,0 +1,257 @@
+import * as assert from 'assert';
+import * as ccc from '../codecoverageconstants';
+import * as expectedResults from './data/expectedResults';
+import * as fakeData from './data/fakeData';
+import * as path from 'path';
+import * as Q from 'q';
```
KonstantinTyukalov, line 6: [Suggestion] Can we try not use the Q library?
It's deprecated and soon should be from task lib.
I believe, basic promises should be backward-compatible with it

PR #148, opened 2023-05-05
Centauri fix

patelchandni:
Fix is applied from:
https://github.com/microsoft/azure-pipelines-tasks/pull/18236

@onetocny


onetocny (approved):
Thanks for your job. Looks nice.

manolerazvan (approved):

PR #152, opened 2023-05-18
Using fileVersion when productVersion is 0.0.0.0

amp-powell:


common-npm-packages/packaging-common/nuget/ProductVersionHelper.ts
```diff
@@ -0,0 +1,12 @@
+import {VersionInfo} from "../pe-parser/VersionResource";
+
+export function getVersionFallback(version: VersionInfo): VersionInfo {
+    const productVersion = version.productVersion;
+    if ((productVersion.a === 0) && (productVersion.b === 0) && (productVersion.c === 0) && (productVersion.d === 0)) {
```
ksigmund, line 5: This might be made more concise by doing something like:

const { a, b, c, d } = productVersion;
if (a === 0 && b === 0 && c === 0 && d === 0) {
  version.productVersion.a = version.fileVersion.a;
  version.productVersion.b = version.fileVersion.b;
  version.productVersion.c = version.fileVersion.c;
  version.productVersion.d = version.fileVersion.d;
}

common-npm-packages/packaging-common/nuget/ProductVersionHelper.ts
```diff
@@ -0,0 +1,12 @@
...
+        version.productVersion.a = version.fileVersion.a;
+        version.productVersion.b = version.fileVersion.b;
+        version.productVersion.c = version.fileVersion.c;
+        version.productVersion.d = version.fileVersion.d;
+    }
+    return version;
```
jmyersmsft, line 11: Instead of altering the passed-in version, you could return the whole `version.productVersion` or `version.fileVersion` object. Generally, callers don't expect functions to modify their arguments, especially when the function's name starts with 'get'.

common-npm-packages/packaging-common/nuget/ProductVersionHelper.ts
amp-powell: Made this change!

magleaso (approved):
LGTM

jmyersmsft (approved):

zhenghao104:
You need to bump the version in package.json and build the package

zhenghao104 (changes requested):
Bump versions in package.json and build the package to update the lock file

zhenghao104 (approved):

PR #157, opened 2023-06-05
Temp nuget.config generated by DotNetCoreCLI@2 has bug

tintse-thxsky-MSFT:
https://github.com/microsoft/azure-pipelines-tasks/pull/17287

### DIFF
--- a/common-npm-packages/packaging-common/package-lock.json
+++ b/common-npm-packages/packaging-common/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "azure-pipelines-tasks-packaging-common",
-    "version": "3.224.0",
+    "version": "3.226.0",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/common-npm-packages/packaging-common/package.json
+++ b/common-npm-packages/packaging-common/package.json
@@ -1,6 +1,6 @@
 {
     "name": "azure-pipelines-tasks-packaging-common",
-    "version": "3.224.0",
+    "version": "3.226.0",
     "description": "Azure Pipelines Packaging Tasks Common",
     "scripts": {
         "test": "echo \"Error: no test specified\" && exit 1",
--- a/common-npm-packages/packaging-common/universal/ArtifactToolUtilities.ts
+++ b/common-npm-packages/packaging-common/universal/ArtifactToolUtilities.ts
@@ -144,6 +144,9 @@ export async function getPackageNameFromId(serviceUri: string, accessToken: stri
     }
 }
 
+/**
+ * Returns the highest package version from a Feed for a given package. If no versions for this package name exist, we will return highest version of 0.0.0
+ */
 export async function getHighestPackageVersionFromFeed(serviceUri: string, accessToken: string, projectId: string, feedId: string, packageName: string): Promise<string> {
     const ApiVersion = "3.0-preview.1";
     const PackagingAreaName = "Packaging";
@@ -175,5 +178,5 @@ export async function getHighestPackageVersionFromFeed(serviceUri: string, acces
         }
     }
     
-    return null;
+    return "0.0.0";
 }


tintse-thxsky-MSFT:
original issue https://github.com/microsoft/azure-pipelines-tasks/pull/17287

common-npm-packages/packaging-common/nuget/NuGetConfigHelper2.ts
```diff
@@ -10,6 +10,10 @@ import * as ngToolRunner from "./NuGetToolRunner2";
 import { NuGetExeXmlHelper } from "./NuGetExeXmlHelper";
 import { NuGetXmlHelper } from "./NuGetXmlHelper";
 import * as ngutil from "./Utility";
+import { getVersionFallback } from "./ProductVersionHelper";
+import peParser = require('../pe-parser/index');
+import * as nuGetGetter from "./NuGetToolGetter";
```
magleaso, line 15: Use these to get the nuget version

common-npm-packages/packaging-common/nuget/NuGetConfigHelper2.ts
```diff
@@ -64,76 +66,105 @@ export class NuGetConfigHelper2 {
...
-    {
+    /**
+ * Set authentication for source in a temporary nuget config file.
+ *
+ *
+ * @param nugetVersion - Optional.  If nugetVersion is provided, will use it to check if version is equal to 4.9.2.  Nuget 4.9.2 has a bug that requires a special work around.  If nugetVersion is not provided, will assusm that nuget version is not 4.9.2.
+ * @returns void. This method set authentication for source in a temporary nuget config file.
+ *
+ */
+    public setAuthForSourcesInTempNuGetConfig(nugetVersion?: VersionInfoVersion): void {
```
magleaso, line 83: Awkward to treat dotnet and nuget 2 different ways. For NuGet, the caller is responsible for resolving the tools version, for dotnet, the callee resolves the versions. Resolve the nuget tools version in this function

common-npm-packages/packaging-common/nuget/NuGetConfigHelper2.ts
```diff
@@ -64,76 +66,105 @@ export class NuGetConfigHelper2 {
...
+                    append = true;
+                    tl.debug(`Dotnet Version ${version} detected.  Appending "feed-" to the key`);
+                }
+
+            } catch (err) {
+                tl.debug(err.message);
+                append = false;
+            }
+        }
+        this.setAuthForSourcesInTempNuGetConfigHelper(sources, append);
```
magleaso, line 117: Instead of factoring this code out and giving it a duplicate name, factor out the rest of the code above and give a more descriptive name like `validateToolsVersions` or something else to denote special handling based on tools versions

magleaso (changes requested):

common-npm-packages/packaging-common/nuget/NuGetConfigHelper2.ts
```diff
@@ -177,6 +149,39 @@ export class NuGetConfigHelper2 {
         });
     }
 
+    private validateToolsVersion() {
```
magleaso, line 152: Specify a return type

common-npm-packages/packaging-common/nuget/NuGetConfigHelper2.ts
```diff
@@ -146,6 +149,39 @@ export class NuGetConfigHelper2 {
...
+        let append = false;
+        let nugetVersion = tl.getInput('versionSpec', false);
```
magleaso, line 154: Is this the correct way to get the nuget version? I see the nuGetGetter import is unused

common-npm-packages/packaging-common/nuget/NuGetConfigHelper2.ts
```diff
@@ -146,6 +149,39 @@ export class NuGetConfigHelper2 {
         });
     }
 
+    private validateToolsVersion() {
+        let append = false;
```
magleaso, line 153: This variable name doesn't mean anything in the context of the function. The idea that we're "appending" something is not relevant to what this function does, simply determining if the tools match a pattern

common-npm-packages/packaging-common/nuget/NuGetConfigHelper2.ts
```diff
@@ -146,6 +149,39 @@ export class NuGetConfigHelper2 {
...
+                let version = dotnet.execSync().stdout.trim();
+                let versionArr = version.split(".");
+                if ((versionArr[0] == "2") && (versionArr[1] == "1") && (versionArr[2] == "500")) {
+                    append = true;
+                    tl.debug(`Dotnet Version ${version} detected.  Appending "feed-" to the key`);
+                }
+
+            } catch (err) {
+                tl.debug(err.message);
+                append = false;
```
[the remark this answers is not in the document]
tintse-thxsky-MSFT: changed the code now that if the nuget version is 4.9.2 then it won't go to dotnet path.
if dotnet get version throws error that I assume that the version is not the 2.1.500 (purely based on probability)

common-npm-packages/packaging-common/nuget/NuGetConfigHelper2.ts
tintse-thxsky-MSFT: I am not too sure tbh. 

1. When I look into NuGetGetter, none of the methods actually return me the currently in use Nuget version.  

2. I tried looking into tl.which('nuget') but that seems to be too many manual code to check.  After getting the nuget exe path,  there is not easy way to get the version. I have to trim off something or grep.
https://stackoverflow.com/questions/15964935/get-nuget-exe-version-from-command-line

3.  So I looked into nugettoolinstaller https://github.com/microsoft/azure-pipelines-tasks/blob/6ea095ec7e00d2c33b91c6a553ef3c587c509233/Tasks/XamariniOSV2/xamarinios.ts#L78 and see that in order to install a specific version, the env variable versionSpec will be set.  So assuming that the env variable versionSpec is **NOT** cleared between tasks' execution, I decided to get the version this way.

magleaso: Looks like other places in the code [call getFileVersionInfoAsync](https://github.com/microsoft/azure-pipelines-tasks/blob/92f8d738551328d50fefa04129fd0fce88456d92/Tasks/Common/packaging-common/pe-parser/index.ts#L12) with [a path to the executable](https://github.com/microsoft/azure-pipelines-tasks/blob/92f8d738551328d50fefa04129fd0fce88456d92/Tasks/NuGetCommandV2/nugetrestore.ts#L76)

jmyersmsft: calling getInput will not do what you want.
* getInput gets an input to the currently executing task instance. Subsequent tasks in the pipeline will not contain that variable
* the versionSpec input in the tasks that take it may be an expression describing a _range_ of acceptable versions (a common one being something like "5.x")

jackbond:
I experience this bug again yesterday. All I can say is, if someone in my organization took this long to fix a bug this simple, they would have been FIRED ages ago.

PR #203, opened 2023-08-02
Xml Variable Specification

DaDarkWizard:
Updated Xml file variable replacement to allow absolute variable specification, fixing a variable collision issue where a connection string and an app setting with the same name could not be updated using pipeline variables in FileTransformV2.

Additional information about this issue can be found in [microsoft/azure-devops-yaml-schema#209](https://github.com/MicrosoftDocs/azure-devops-yaml-schema/issues/209)

Variables can now use dot syntax to specify which setting to update.

appSettings.var1 will update var1 in appSettings
connectionStrings.var1 will update var1 in connectionStrings

Variables will still behave as used previously unless the more precise one is given.


DaDarkWizard:

PR #216, opened 2023-08-24
Fix finding latest version of new package

AlexVTor:
This PR contains changes to return a correct package version (0.0.0) when publishing a upack with a new unique name

### DIFF
--- a/common-npm-packages/packaging-common/universal/ArtifactToolUtilities.ts
+++ b/common-npm-packages/packaging-common/universal/ArtifactToolUtilities.ts
@@ -175,5 +175,5 @@ export async function getHighestPackageVersionFromFeed(serviceUri: string, acces
         }
     }
     
-    return null;
+    return "0.0.0";
 }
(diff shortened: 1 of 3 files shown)


common-npm-packages/packaging-common/universal/ArtifactToolUtilities.ts
```diff
@@ -175,5 +175,5 @@ export async function getHighestPackageVersionFromFeed(serviceUri: string, acces
         }
     }
     
-    return null;
+    return "0.0.0";
```
AlexVTor, line 178: This function is only used by the universalpackagesV0 task. As the function name implies, the highestPackageVersion should be returned. In the event no package yet exists, a (0.0.0) should be returned.

jmyersmsft (approved):

magleaso (changes requested):
Can we add a doc string comment on the getHighestPackageVersionFromFeed to clarify what the expected return value is in each scenario and possibly also a comment on the caller

zhenghao104 (approved):

magleaso (approved):
LGTM

PR #228, opened 2023-10-16
Token authentication in MSDeploy

onetocny:
This PR adds token authentication support to `azure-pipelines-tasks-webdeployment-common` library. Changes are done in two major directions:
- Possibility to pass authentication type as an argument to MSDeploy,
- Provide MSDeploy version that supports token atuhentication.

### DIFF
--- a/common-npm-packages/webdeployment-common/Tests/L0MSDeployUtility.ts
+++ b/common-npm-packages/webdeployment-common/Tests/L0MSDeployUtility.ts
@@ -32,6 +32,18 @@ else {
     throw new Error('MSBUILD PACKAGE DEFAULT PARAMS FAILED');
 }
 
+var tokenAuthMSBuildPackageArgument: string = msdeployUtility.getMSDeployCmdArgs('package.zip', 'webapp_name', {
+    publishUrl: 'publishUrl', userName: 'user', userPWD: 'token'
+}, true, false, true, null, null, null, true, false, false, "Bearer");
+
+console.log(` * MSBUILD TOKEN AUTH PARAMS: ${tokenAuthMSBuildPackageArgument}`);
+if(checkParametersIfPresent(tokenAuthMSBuildPackageArgument, ["AuthType=\"'Bearer'\""])) {
+    console.log("MSBUILD TOKEN AUTH PARAMS PASSED");
+}
+else {
+    throw new Error('MSBUILD TOKEN AUTH PARAMS FAILED');
+}
+
 
 var packageWithSetParamArgument: string = msdeployUtility.getMSDeployCmdArgs('package.zip', 'webapp_name', {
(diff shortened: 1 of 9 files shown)


starkmsu (approved):

common-npm-packages/webdeployment-common/Tests/L0MSDeployUtility.ts
```diff
@@ -32,6 +32,18 @@ else {
     throw new Error('MSBUILD PACKAGE DEFAULT PARAMS FAILED');
 }
 
+var tokenAuthMSBuildPackageArgument: string = msdeployUtility.getMSDeployCmdArgs('package.zip', 'webapp_name', {
+    publishUrl: 'publishUrl', userName: 'user', userPWD: 'token'
+}, true, false, true, null, null, null, true, false, false, "Bearer");
```
LeftTwixWand, line 37: I can suggest using named arguments for readability, but it's not mandatory, so feel free to resolve this comment.

common-npm-packages/webdeployment-common/msdeployutility.ts
```diff
@@ -206,82 +207,156 @@ function parseAdditionalArguments(additionalArguments: string): string[] {
...
+            if (!installPathItem) {
+                reject(tl.loc("MissingMSDeployInstallPathRegistryKey"));
+            }
+
+            resolve(installPathItem.value);
+        });
     });
+}
+
+function compareVersions(version1: string, version2: string): number {
```
LeftTwixWand, line 330: Can we replace this method with https://github.com/semver/semver ?
Like this:
```js
import * as semver from 'semver';

const result = semver.compare('1.2.3', '1.2.4');
```

common-npm-packages/webdeployment-common/Tests/L0MSDeployUtility.ts
onetocny: How can we do that in TypeScript? I was under impression that there is not support for that.

common-npm-packages/webdeployment-common/msdeployutility.ts
onetocny: Yeah, I was considering that. But it would require to add additional reference to that package. Since we really need basic versions comparison I decided to do it directly by our code.

LeftTwixWand: You're right

common-npm-packages/webdeployment-common/Tests/L0MSDeployUtility.ts
LeftTwixWand: You're right

LeftTwixWand (approved):

PR #232, opened 2023-10-18
Transform webdeployment-common tests into real test suites

onetocny:
Until now tests from `webdeployment-common` were not run during package build process. Instead of that pipeline tasks that are using this package were running JavaScript files from `webdeployment-common\Tests` folder. Also there were missing real test suites (`describe`) with particular tests (`it`). This PR introduces new `L0.ts` file that contains calls to all existing tests. All tests were wrapped into test suites and additional fixes were applied where it was necessary.

### DIFF
--- a/common-npm-packages/webdeployment-common/Tests/L0CopyDirectory.ts
+++ b/common-npm-packages/webdeployment-common/Tests/L0CopyDirectory.ts
@@ -1,5 +1,6 @@
 import * as assert from "assert";
 import * as mockery from "mockery";
+import * as path from "path";
 
 export function runCopyDirectoryTests(): void {
     const fileList: string[] = [];
@@ -73,16 +74,16 @@ export function runCopyDirectoryTests(): void {
 
     it("Should copy files and folders as expected", async () => {
         fileList.push(
-            "C:\\source\\path",
-            "C:\\source\\path\\myfile.txt",
-            "C:\\source\\path\\New Folder",
-            "C:\\source\\path\\New Folder\\Another New Folder",
-            "C:\\source\\New Folder\\anotherfile.py",
-            "C:\\source\\New Folder\\Another New Folder\\mynewfile.txt"
+            path.join("C:", "source", "path"),
+            path.join("C:", "source", "path", "myfile.txt"),
+            path.join("C:", "source", "path", "New Folder"),
+            path.join("C:", "source", "path", "New Folder", "Another New Folder"),
+            path.join("C:", "source", "New Folder", "anotherfile.py"),
+            path.join("C:", "source", "New Folder", "Another New Folder", "mynewfile.txt")
         );
         const utility = await import('../utility');
 
-        utility.copyDirectory('C:\\source', 'C:\\destination');
+        utility.copyDirectory(path.join('C:','source'), path.join('C:', 'destination'));
 
         assert.strictEqual(cpfilesCount, 3, '## Copy Files Successful ##');
         assert.strictEqual(mkdirPCount, 6, '## mkdir Successful ##');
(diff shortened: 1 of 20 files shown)


common-npm-packages/webdeployment-common/Tests/L0CopyDirectory.ts
```diff
@@ -1,59 +1,91 @@
...
+                    return;
+                }
+                cpfilesCount++;
+                fileList.push(dest);
+                console.log('cp ' + source + ' to ' + dest);
+            },
+            stats: function (path: string): any {
+                return {
+                    isDirectory: function (): boolean {
+                        return !path.endsWith('.py') && !path.endsWith('.txt');
```
VictorGavrish, line 44: That's an interesting way to determine if something is a directory :) Maybe use types instead? Instead of using strings, we could have `fileList: (File | Directory)[]` or something like that?

common-npm-packages/webdeployment-common/utility.ts
```diff
@@ -1,10 +1,8 @@
 import path = require('path');
-import fs = require('fs');
 import * as os from "os";
 import tl = require('azure-pipelines-task-lib/task');
-import { PackageUtility, PackageType } from './packageUtility';
-
-var zipUtility = require('azure-pipelines-tasks-webdeployment-common/ziputility.js');
+import {  PackageType } from './packageUtility';
```
VictorGavrish, line 4: extra space?

common-npm-packages/webdeployment-common/Tests/L0CopyDirectory.ts
onetocny: Fixed.

common-npm-packages/webdeployment-common/utility.ts
onetocny: Fixed

VictorGavrish (approved):
LGTM

common-npm-packages/webdeployment-common/packageUtility.ts
```diff
@@ -67,7 +67,7 @@ export class Package {
 
     public async isMSBuildPackage(): Promise<boolean> {
```
LeftTwixWand, line 68: Can we refactor this method to add a bit more clarity?
```suggestion
    public async isMSBuildPackage(): Promise<boolean> {
  if (this._isMSBuildPackage !== undefined) return this._isMSBuildPackage;

  const shouldCheckFiles = this.getPackageType() !== PackageType.folder;
  this._isMSBuildPackage = shouldCheckFiles && await checkIfFilesExistsInZip(this._path, ["parameters.xml", "systeminfo.xml"]);

  tl.debug(`Is the package an msdeploy package : ${this._isMSBuildPackage}`);

  return this._isMSBuildPackage;
}
```

common-npm-packages/webdeployment-common/Tests/L1XmlVarSub.ts
```diff
@@ -1,29 +1,86 @@
...
+import * as fs from 'fs';
+
+import * as tl from 'azure-pipelines-task-lib';
+
+import { substituteXmlVariables } from '../xmlvariablesubstitutionutility';
+import { detectFileEncoding } from "../fileencoding";
+
+export function runL1XmlVarSubTests(this: Mocha.Suite): void {
+
+    this.timeout(60000);
```
LeftTwixWand, line 13: Do we really need this timeout, because from what I see - this test is synchronous.

LeftTwixWand (commented):
Looks good👍👍

common-npm-packages/webdeployment-common/Tests/L1XmlVarSub.ts
onetocny: I think the `timeout` is applicable to sync tests too, isn't it? In this particular case it means that whole suite has to finish under one minute. Otherwise it fails.

PR #241, opened 2023-10-27
Expose function to get version of installed MSDeploy

onetocny:
Web deployment tasks need to detect if there is a MSDeploy that supports token authentication installed on the agent machine. In opposite the task should not start the deployment at all and fail instantly. In order to do so this PR introduces `installedMSDeployVersionSupportsTokenAuth` that provides such information. More details in [work item](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2118491).

### DIFF
--- a/common-npm-packages/webdeployment-common/msdeployutility.ts
+++ b/common-npm-packages/webdeployment-common/msdeployutility.ts
@@ -1,14 +1,11 @@
-import Q = require('q');
 import tl = require('azure-pipelines-task-lib/task');
-import trm = require('azure-pipelines-task-lib/toolrunner');
 import fs = require('fs');
 import path = require('path');
-import { Package, PackageType } from './packageUtility';
+import { Package } from './packageUtility';
+import * as winreg from 'winreg';
+import * as semver from 'semver';
 
-var winreg = require('winreg');
-var parseString = require('xml2js').parseString;
-const ERROR_FILE_NAME = "error.txt";
-
+export const ERROR_FILE_NAME = "error.txt";
 /**
  * Constructs argument for MSDeploy command
  * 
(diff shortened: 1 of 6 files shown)


common-npm-packages/webdeployment-common/msdeployutility.ts
```diff
@@ -254,111 +250,91 @@ export async function getMSDeployFullPath(): Promise<string> {
...
+                if (!isNaN(subKeyVersionNumber) && subKeyVersionNumber > latestKeyVersion) {
+                    latestKeyVersion = subKeyVersionNumber;
+                    latestSubKey = subRegKeys[index];
                 }
             }
-
-            const installPathItem = items.find(item => item.name === "InstallPath");
-            if (!installPathItem) {
-                reject(tl.loc("MissingMSDeployInstallPathRegistryKey"));
+            if (latestKeyVersion < 3) {
```
VictorGavrish, line 329: Why does the key have to be less than 3? I somehow lost the logic of this piece of code.

onetocny: This logic is iterating through all installed versions of msdeploy in registry. Details about every version are stored under separate sub key (see below). In my case I have only one version. But in case there are multiple versions the highest one which is greater than 3 should be picked. I believe that versions older than 3 are not compatible either with our tasks or with Azure App Service API. 

image

VictorGavrish: Okay, please elaborate on this somewhere in comments, otherwise it's impossible to know the logic behind this code in the future.

onetocny: Fixed. Thank you for your review.

VictorGavrish (approved):
LGTM

PR #250, opened 2023-11-13
Remove extra part of blob url

aleksandrlevochkin:
`blockBlobClient` appears to hold the full blob url internally.


aleksandrlevochkin:
Abandoned in favor of https://github.com/microsoft/azure-pipelines-tasks-common-packages/pull/258

PR #268, opened 2024-01-24
Fix CG in Azure-blobstorage-provider

KonstantinTyukalov:
WI: [2137579](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2137579)

Tested with JavaToolInstaller@1 task - unit tests + [e2e run](https://abtttestorg.visualstudio.com/JavaToolInstallerTest/_build/results?buildId=7128)

### DIFF
--- a/common-npm-packages/az-blobstorage-provider/package.json
+++ b/common-npm-packages/az-blobstorage-provider/package.json
@@ -1,6 +1,6 @@
 {
     "name": "azp-tasks-az-blobstorage-provider",
-    "version": "3.232.3",
+    "version": "3.235.0",
     "description": "Common Library to download blobs from azure storage",
     "repository": {
         "type": "git",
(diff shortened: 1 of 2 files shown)


common-npm-packages/az-blobstorage-provider/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "azp-tasks-az-blobstorage-provider",
-    "version": "3.232.2",
+    "version": "3.232.3",
```
aleksandrlevochkin, line 3: Should not the version be 3.235.0?

aleksandrlevochkin (approved):

common-npm-packages/az-blobstorage-provider/package.json
KonstantinTyukalov: It's not the task, so I'm not sure it makes sense to keep the sprint number as version for common packages.
I'll have a look how we take versioning for other packages and return back

KonstantinTyukalov: ok, seems others follow the same rule as for tasks. I won't break this consistency in this case. Bumped to 3.235.0

Pranshu-Negi (approved):

kirill-ivlev (approved):

DergachevE (approved):

vmapetr (approved):

PR #290, opened 2024-02-29
[AzureRMDeployCommon] Replace decompress-zip with node-stream-zip

kirill-ivlev:
**Description**:
This PR replacing `decompress-zip` package with `node-stream-api` for `AzureRMDeployCommon package since decompress-zip have incompatibility issue with Node17+

Links to WI:
https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2150886
https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2150886

### DIFF
--- a/common-npm-packages/azurermdeploycommon/webdeployment-common/ziputility.ts
+++ b/common-npm-packages/azurermdeploycommon/webdeployment-common/ziputility.ts
@@ -13,6 +13,7 @@ export async function unzip(zipLocation, unzipLocation) {
     tl.mkdirP(unzipLocation);
     const zip = new StreamZip.async({ file: zipLocation});
         zip.extract(null, unzipLocation).then(() => {
+            zip.close();
             defer.resolve();
         }).catch((error) => {
             console.log(`error extracting ${zipLocation}: ${error}`);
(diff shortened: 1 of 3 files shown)


manolerazvan (approved):

common-npm-packages/azurermdeploycommon/webdeployment-common/ziputility.ts
```diff
@@ -2,28 +2,23 @@ import tl = require('azure-pipelines-task-lib/task');
...
-    unzipper.on('extract', function (log) {
-        tl.debug('extracted ' + zipLocation + ' to ' + unzipLocation + ' Successfully');
-        defer.resolve(unzipLocation);
-    });
-    unzipper.extract({
-      path: unzipLocation
-    });
-    return defer.promise;
+    tl.mkdirP(unzipLocation);
+    const zip = new StreamZip.async({ file: zipLocation});
```
merlynomsft, line 14: [Not blocking as I'm sure it's well-tested]:  but it would be nice if the new code path were switched using a parameter so we can fall back to the old code path for node16.  E.g. if node20 + ode-stream-zip has a regression, we can use the buildconfig to revert to old code path.

merlynomsft (approved):

common-npm-packages/azurermdeploycommon/webdeployment-common/ziputility.ts
kirill-ivlev: @merlynomsft we can downgrade version of this package in case if we will face problem, and rollback changes in the next version of common package

PR #316, opened 2024-05-26
Don't try to read a null response when getting a federated token

50Wliu:
If response is null on L68, you can bet response.oidcToken is going to throw on L72...


tintse-thxsky-MSFT:
LGTM

tintse-thxsky-MSFT (approved):

50Wliu:
Thanks for the review! Can someone also merge this please? :)

PR #327, opened 2024-06-17
Package Common Npm Bearer Auth Mask

tintse-thxsky-MSFT:


common-npm-packages/packaging-common/npm/npmregistry.ts
```diff
@@ -25,6 +25,7 @@ export class NpmRegistry implements INpmRegistry {
         this.authOnly = authOnly || false;
     }
 
+    /** Return NpmRegistry with masked auth from Service Endpoint. */
```
embetten, line 28: nit: why multiline comment when its a single line?

common-npm-packages/packaging-common/npm/npmregistry.ts
```diff
@@ -59,31 +60,31 @@ export class NpmRegistry implements INpmRegistry {
                 username = endpointAuth.parameters['username'];
                 password = endpointAuth.parameters['password'];
                 email = username; // npm needs an email to be set in order to publish, this is ignored on npmjs
-                password64 = (new Buffer(password).toString('base64'));
```
embetten, line 62: why move the secret call from here to below? is this code not working?

common-npm-packages/packaging-common/npm/npmregistry.ts
```diff
@@ -59,31 +60,31 @@ export class NpmRegistry implements INpmRegistry {
...
                 const apitoken = endpointAuth.parameters['apitoken'];
+                tl.setSecret(apitoken);
                 if (!isVstsTokenAuth){
                     // Use Bearer auth as it was intended.
                     auth = nerfed + ':_authToken=' + apitoken + lineEnd;
                 } else {
                     // Azure DevOps does not support PATs+Bearer only JWTs+Bearer
                     email = 'VssEmail';
                     username = 'VssToken';
-                    password64 = (new Buffer(apitoken).toString('base64'));
```
embetten, line 78: should this password field be entirely removed ?

common-npm-packages/packaging-common/npm/npmregistry.ts
tintse-thxsky-MSFT: I think this is the syntax for method docs.

common-npm-packages/packaging-common/npm/npmregistry.ts
tintse-thxsky-MSFT: just refactoring the secret code outside of switch block. cleaner code.

common-npm-packages/packaging-common/npm/npmregistry.ts
tintse-thxsky-MSFT: oops I accidently deleted it, added back,

common-npm-packages/packaging-common/npm/npmregistry.ts
```diff
@@ -59,31 +60,32 @@ export class NpmRegistry implements INpmRegistry {
...
-                    tl.setSecret(password64);
+                    password64 = Buffer.from(apitoken).toString('base64');
 
                     auth = nerfed + ':username=' + username + lineEnd;
                     auth += nerfed + ':_password=' + password64 + lineEnd;
                     auth += nerfed + ':email=' + email + lineEnd;
                 }
                 break;
         }
+        tl.setSecret(password);
```
AlexVTor, line 87: From the code above, only one of these lines will actually be used in any scenario. Can we use the .setSecret() method on an object that hasn't been set?

tintse-thxsky-MSFT: yea the function setSecret does do a null/empty string check before doing anything.

/**
 * Registers a value with the logger, so the value will be masked from the logs.  Multi-line secrets are not allowed.
 *
 * @param val value to register
 */
function setSecret(val) {
    if (val) {
        if (val.match(/\r|\n/) && ("" + process.env['SYSTEM_UNSAFEALLOWMULTILINESECRET']).toUpperCase() !== 'TRUE') {
            throw new Error(exports.loc('LIB_MultilineSecret'));
        }
        exports.command('task.setsecret', {}, val);
    }
}

AlexVTor (approved):

PR #330, opened 2024-06-18
Artifacts-Common Added setseceret in constructor and abstract method in Service Connection

tintse-thxsky-MSFT:


tintse-thxsky-MSFT:
I have to add mockery as dependency because our test cases are using it since the beginning.  Don't know why/how it got removed.

embetten:
> I have to add mockery as dependency because our test cases are using it since the beginning. Don't know why/how it got removed.

Mockery was removed because it has a critical vulnerability and is no longer maintained. Adding it back is not an option

embetten (dismissed):
remove mockery.

tintse-thxsky-MSFT:
changing it to import * as mocker from "azure-pipelines-task-lib/lib-mocker";

jmyersmsft (approved):
doing "work" in a constructor isn't my favorite thing, but I don't want to block this on that.

PR #342, opened 2024-07-02
Add Packaging WIF Handshake for Packaging Authenticate Tasks

embetten:
- expose get feed tenant id function 
- expose function to get federated token


common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+import * as tl from 'azure-pipelines-task-lib/task';
+import { getSystemAccessToken } from "./webapi";
+import fetch from "node-fetch";
+
+tl.setResourcePath(path.join(__dirname, 'module.json'), true);
+
+export async function getFederatedWorkloadIdentityCredentials(serviceConnectionName: string, tenantId?: string) : Promise<string>
+{
+    try {
+        let tenant = tenantId ?? tl.getEndpointAuthorizationParameterRequired(serviceConnectionName, "TenantId");
```
cormacpayne, line 11: nit: do we need to handle the case where `tl.getEndpointAuthorizationParameterRequired` does not find a value for `TenantId` in the service connection? I'm not sure if that's possible based on the structure of a service connection in this particular case, but wanted to double-check before we start using the result in places.

common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+        let entraURI = "https://login.windows-ppe.net/"+tenant+"/oauth2/v2.0/token";
+        let clientId = tl.getEndpointAuthorizationParameterRequired(serviceConnectionName, "ServicePrincipalId");
+
+        let body = {
+            'scope': "499b84ac-1321-427f-aa17-267ca6975798/.default",
+            'client_id': clientId,
+            'client_assertion_type': "urn:ietf:params:oauth:client-assertion-type:jwt-bearer",
+            'client_assertion': ADOResponse.oidcToken,
+            'grant_type': "client_credentials"
+        };
```
cormacpayne, line 36: nit: do you feel it's worth moving the hard-coded values in this body to a constants file? I know something like the `scope` you've provided is the resource for Azure DevOps when acquiring a token, so that might be something worth explicitly specifying in a constants file and sharing across other auth common code here? (similar comment may apply to the other hard-coded values, I don't know how expansive their scenarios are)

common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+            method: 'POST', 
+            headers: 
+            {
+                'Content-Type': 'application/json',
+                'Authorization': 'Bearer '+ systemAccessToken
+            }
+        })).json() as {oidcToken: string};
+
+        tl.setSecret(ADOResponse.oidcToken);
+        let entraURI = "https://login.windows-ppe.net/"+tenant+"/oauth2/v2.0/token";
```
cormacpayne, line 27: nit: just verifying that we should be hitting the PPE endpoint here

common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+                'Content-Type': 'application/x-www-form-urlencoded'
+            }
+        })).json() as {access_token: string};
+
+        tl.setSecret(entraResponse.access_token);
+        return entraResponse.access_token;
+    } 
+    catch (error) 
+    {
+        tl.error(tl.loc("Error_FederatedTokenAquisitionFailed", error));
```
cormacpayne, line 57: nit: do we need to return a `string` in this `catch` block? We're catching an error and writing it to the error stream, but that doesn't terminate the method call, right? I figured we would still need to return a `string` from this method block, but I may be wrong 🙂

common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+export async function getFeedTenantId(feedUrl: string) : Promise<string>
+{
+    try
+    {
+        const feedResponse =  await fetch(feedUrl);
+        return feedResponse?.headers?.get('X-VSS-ResourceTenant');
+    } 
+    catch (error)
+    {
+        tl.warning(tl.loc("Error_GetFeedTenantIdFailed", error));
```
cormacpayne, line 70: nit: similar comment here -- do we need to return a `string` in this `catch` block?

cormacpayne (approved):
Just some clarifying questions, otherwise LGTM :shipit:

common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
embetten: grr we should not brb

common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
embetten: The underlying function getEndpointAuthorizationParameterRequired  throws if the required filed is not found. So as long as we are handling the catch correctly, it should already be covered

common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
embetten: I really dont expect these to be used anywhere else but they are somewhat magic strings

common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
embetten: fixed

common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+        const entraResponse: {access_token: string} = await (await fetch(entraURI, 
+        {
+            method: 'POST', 
+            body: formBody,
+            headers: 
+            {
+                'Content-Type': 'application/x-www-form-urlencoded'
+            }
+        })).json() as {access_token: string};
+        tl.debug("access token: " + entraResponse.access_token)
```
cormacpayne, line 50: nit: should we be writing this to any stream? 😳

embetten: should we? how would we use a stream here if are deserializing the response to the `{access_token: string}` object ?

common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
embetten: Yes ! i updated the catch to return undefined and updated the return type to explicitly call out undefined

common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts
embetten: same as above :)

PR #354, opened 2024-07-12
Fix typo in ArtifactToolsUtilities

MicroTuld:
Fixes a typo -> Artifacttool to ArtifactTool

### DIFF
--- a/common-npm-packages/packaging-common/package.json
+++ b/common-npm-packages/packaging-common/package.json
@@ -1,6 +1,6 @@
 {
     "name": "azure-pipelines-tasks-packaging-common",
-    "version": "3.241.1",
+    "version": "3.242.1",
     "description": "Azure Pipelines Packaging Tasks Common",
     "scripts": {
         "test": "mocha _build/Tests/L0.js",
--- a/common-npm-packages/packaging-common/package.json
+++ b/common-npm-packages/packaging-common/package.json
@@ -1,6 +1,6 @@
 {
     "name": "azure-pipelines-tasks-packaging-common",
-    "version": "3.242.1",
+    "version": "3.242.0",
     "description": "Azure Pipelines Packaging Tasks Common",
     "scripts": {
         "test": "mocha _build/Tests/L0.js",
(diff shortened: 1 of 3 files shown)


common-npm-packages/packaging-common/universal/ArtifactToolUtilities.ts
```diff
@@ -11,7 +11,7 @@ import * as toollib from "azure-pipelines-tool-lib/tool";
 export function getArtifactToolLocation(dirName: string): string {
     let toolPath: string = path.join(dirName, "ArtifactTool.exe");
     if (tl.osType() !== "Windows_NT"){
-        toolPath = path.join(dirName, "Artifacttool");
+        toolPath = path.join(dirName, "ArtifactTool");
```
jmyersmsft, line 14: You mentioned offline that a previous version of the task worked, I believe with the same version of ArtifactTool that was being used in the failing instance, with `artifacttool` all lowercase. If the file is all lowercase in the zip file, it needs to be all lowercase here too.

common-npm-packages/packaging-common/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "azure-pipelines-tasks-packaging-common",
-    "version": "3.241.1",
+    "version": "3.242.1",
```
embetten, line 3: 3.242.0 ?

MicroTuld: The final digit represents the "Count of changes was made during the sprint" -- so that means it must always be > 0?
https://mseng.visualstudio.com/AzureDevOps/_wiki/wikis/AzureDevOps.wiki/29720/Mitigate-the-issue-Override-task-version

common-npm-packages/packaging-common/universal/ArtifactToolUtilities.ts
MicroTuld: That's a bit of a conundrum, though. Zheng was moving to `ArtifactTool` because other uses were complaining that all-lowercase was also not working.

MicroTuld: https://github.com/microsoft/azure-pipelines-tasks/pull/18143

common-npm-packages/packaging-common/package.json
MicroTuld: my interpretation did not hold up to observed practice, so going with a zero here

common-npm-packages/packaging-common/universal/ArtifactToolUtilities.ts
MicroTuld: moving to all lowercase for that path segment

jmyersmsft (approved):

PR #355, opened 2024-07-12
CredProvider - Append singular service connection per run

AlexVTor:
This PR contains work to support the re-runnability of the NuGetAuthenticateV1 task by
- Enforcing the use of a singular service connection in the task
- Appending the new service connection to the existing set of service connections credentials 
- Preventing the duplication of entries in the list


common-npm-packages/artifacts-common/credentialProviderUtils.ts
```diff
@@ -117,6 +117,23 @@ export async function configureCredProvider(protocol: ProtocolType, serviceConne
...
+ * Configure the credential provider to provide credentials for feeds within the pipeline's organization
+ * and for a single provided service connections.
+ */
+export async function configureSingleCredProvider(
+  protocol: ProtocolType,
+  serviceConnections: ServiceConnection[]
+) {
+  await configureCredProviderForSameOrganizationFeeds(protocol);
+
+  if (serviceConnections && serviceConnections.length > 1) {
```
AlexVTor, line 130: I decided to enforce the singular aspect here, rather than cluttering the task. Will this be fine?

common-npm-packages/artifacts-common/credentialProviderUtils.ts
```diff
@@ -155,6 +172,42 @@ export function configureCredProviderForServiceConnectionFeeds(serviceConnection
...
+
+        const externalFeedEndpointsJson = buildExternalFeedEndpointsJson(serviceConnections);
+        const additionalCredentials = JSON.parse(externalFeedEndpointsJson).endpointCredentials;
+        const existingCredentials = configuredEndpoints ? JSON.parse(configuredEndpoints).endpointCredentials : [];
+        
+        for (const cred of existingCredentials) {
+            if (cred.hasOwnProperty("endpoint") && cred["endpoint"] === additionalCredentials[0]["endpoint"]) 
+            { 
+                // 
+                throw Error(tl.loc('CredProvider_Error_ServiceConnectionExists'));
```
AlexVTor, line 198: I threw here, but should the task warn instead of fail here?

common-npm-packages/artifacts-common/package.json
```diff
@@ -23,6 +23,6 @@
     },
     "devDependencies": {
         "@types/node-fetch": "^2.6.11",
-        "typescript": "4.0.2"
```
embetten, line 26: We might need to ask the pipeline's team if upgrading typescript will break node 16 tasks? and if so, how does that work with buildgen and node 20 FF.

common-npm-packages/artifacts-common/module.json
```diff
@@ -10,6 +10,8 @@
         "CredProvider_InstallingNetCoreTo_NuGetAuthenticateV0": "Installing the Azure Artifacts Credential Provider (.NET Core) to '%s'. This credential provider is compatible with dotnet SDK 2.1 or later.",
         "CredProvider_InstallingNetCoreTo_NuGetAuthenticateV1": "Installing the Azure Artifacts Credential Provider (.NET Core) to '%s'. This credential provider is compatible with .NET SDK 6 or later.",
         "CredProvider_InstallingNetFxTo": "Installing the Azure Artifacts Credential Provider (.NET Framework) to '%s'. This credential provider is compatible with nuget.exe 4.8.0.5385 or later, and MSBuild 15.8.166.59604 or later.",
+        "CredProvider_Error_MultipleServiceConnections": "Multiple Service connections are not supported. Only one service connection should be used per task execution.",
+        "CredProvider_Error_ServiceConnectionExists" : "An existing Service connection already exists for this feed.",
```
embetten, line 14: For the Tasks repo, we learned from the last task updates that we need to include the en-us string to allow the localization pipeline to update the other localization.

common-npm-packages/artifacts-common/credentialProviderUtils.ts
```diff
@@ -155,6 +172,42 @@ export function configureCredProviderForServiceConnectionFeeds(serviceConnection
...
+ * Configure the credential provider to provide credentials for a single service connections,
+ * using VSS_NUGET_EXTERNAL_FEED_ENDPOINTS to do so.
+ */
+export function configureSingleCredProviderForServiceConnectionFeeds(serviceConnections: ServiceConnection[]) {
+    var configuredEndpoints = tl.getVariable(CRED_PROVIDER_EXTERNAL_ENDPOINTS_ENVVAR);
+
+    if (serviceConnections) {
+        console.log(tl.loc('CredProvider_SettingUpForServiceConnections'));
+        
+        serviceConnections.map(authInfo => `${authInfo.packageSource.uri}`).forEach(serviceConnectionUri => console.log('  ' + serviceConnectionUri))
```
embetten, line 185: Note, this will work for the current service connection types, but for the new types there is no authInfo.packageSource.Uri so the add single will pass the service connection and feed url separately.

common-npm-packages/artifacts-common/credentialProviderUtils.ts
embetten: Just logging a warning message might hide the issue, but throwing an error would not allow any authenitcation to be added. I think we want the latter behavior and throwing would let the calling tasks catch and handle the error.

embetten: As we discussed offline setting the task result to warntask might not be appropriate from a library. Throwing and catching in the calling tasks might ?

common-npm-packages/artifacts-common/credentialProviderUtils.ts
embetten: this is fine for now, but consider generalizing so we dont have to update in the future

common-npm-packages/artifacts-common/credentialProviderUtils.ts
AlexVTor: This is a good idea. I have throws in here and will catch them in the task PR

common-npm-packages/build-scripts/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "build-scripts",
```
embetten, line 2: should this file be updated ?

common-npm-packages/artifacts-common/credentialProviderUtils.ts
```diff
@@ -114,6 +114,12 @@ export function getUserProfileNuGetPluginsDir(): string {
  */
 export async function configureCredProvider(protocol: ProtocolType, serviceConnections: ServiceConnection[]) {
     await configureCredProviderForSameOrganizationFeeds(protocol);
+
+    // Enforce a singular service connection for the Entra service connection type
+    if (serviceConnections && serviceConnections.length > 1 && serviceConnections[0].authType === ServiceConnectionAuthType.Entra) {
```
embetten, line 119: nit: is there an implementation reason to enforce single service connections here in the and library not the task ? Or if it needs to be in the library, consider a second method that only allows a single service connection instead of an array.

common-npm-packages/artifacts-common/serviceConnectionUtils.ts
```diff
@@ -87,12 +87,28 @@ export class PrivateKeyServiceConnection extends ServiceConnection {
     }
 }
 
+export class EntraServiceConnection extends ServiceConnection {
+    constructor(
+        public packageSource: IExternalPackageSource,
+        public entraConnectionName: string,
+        public token: string,
```
embetten, line 94: The token will not actually be in the the service connection. The Entra wif service connection only holds the MI /SP informatation. The method here: https://github.com/microsoft/azure-pipelines-tasks-common-packages/blob/3d9a0fbbb26f932bf0498f6c06113592bbea1878/common-npm-packages/artifacts-common/EntraWifUserServiceConnectionUtils.ts#L12 actually uses the MI/SP information to do the WIF token handshake to get the token. So for the entra types, you will need to call the above function to actually get the token

common-npm-packages/artifacts-common/credentialProviderUtils.ts
AlexVTor: Made changes to support the new service connection and (hopefully) large avoid future updates

common-npm-packages/build-scripts/package-lock.json
AlexVTor: I don't think so. I reverted the changes and built the task just fine

common-npm-packages/artifacts-common/module.json
AlexVTor: done

common-npm-packages/artifacts-common/package.json
AlexVTor: The suggestion here was to set the version to match what's being used by the node20 task versions. Node16 versions 'shouldnt' break

embetten (approved):
LGTM

PR #360, opened 2024-08-05
support cluster credentials api

smiklos:
The accessProfile API is going to be deprecated not to mention that there's no out of the box RBAC role that permits this API without adding Admin access to the cluster/namespace. 

The newer API differentiates between admin and user credentials and is covered by all built in AKS RBAC roles.

HelmDeployV0 and HelmDeployV1 should be migrated to use this newer API to simplify working with AKS.

### DIFF
--- a/common-npm-packages/azure-arm-rest/azure-arm-aks-service.ts
+++ b/common-npm-packages/azure-arm-rest/azure-arm-aks-service.ts
@@ -28,10 +28,10 @@ export class AzureAksService {
         this._client = new ServiceClient(endpoint.applicationTokenCredentials, endpoint.subscriptionID, 30);
     }
 
-    public beginRequest(uri: string,  parameters: {}) : Promise<webClient.WebResponse> {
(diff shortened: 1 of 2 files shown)


smiklos:

smiklos:
Added some tests but as far as I can tell none of the L0-* tests are running under azure-arm-rest package...

smiklos:
@manolerazvan Could please review this?

smiklos:
@starkmsu any chance you could take a look? this Is related to https://github.com/microsoft/azure-pipelines-tasks/issues/16614


I'm trying to add support for the new APIs without removing the old one for now. If this gets merge we can change HElmDeploy and other tasks to use the newer APIs

YevheniiKholodkov (approved):
Looks good to me. Approved.

smiklos:
Thanks for the review! Are we not concerned that the tests are not running for this package?

smiklos:
ZzzZZz

PR #369, opened 2024-08-21
artifacts-common Revert Nuget Service Connection Overwrite

embetten:
Ensure nuget service connections allow to overwrite when an endpoint has been previously set.

### DIFF
--- a/common-npm-packages/artifacts-common/credentialProviderUtils.ts
+++ b/common-npm-packages/artifacts-common/credentialProviderUtils.ts
@@ -162,40 +162,27 @@ export async function configureCredProviderForSameOrganizationFeeds(protocol: Pr
  * using VSS_NUGET_EXTERNAL_FEED_ENDPOINTS to do so.
  */
 export function configureCredProviderForServiceConnectionFeeds(serviceConnections: ServiceConnection[]) {
-    var configuredEndpoints = tl.getVariable(CRED_PROVIDER_EXTERNAL_ENDPOINTS_ENVVAR);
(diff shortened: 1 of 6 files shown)


AlexVTor (approved):
LGTM

embetten:
> LGTM

This is not quite right - brb

embetten (commented):
Currently appending to array, does not properly replace element

common-npm-packages/artifacts-common/credentialProviderUtils.ts
```diff
@@ -161,48 +161,34 @@ export async function configureCredProviderForSameOrganizationFeeds(protocol: Pr
...
+    var newConnections = buildExternalFeedEndpoints(serviceConnections);
+    var map = new Map(newConnections.map(item => [item.endpoint, item]));
+    const configuredEndpoints = tl.getVariable(CRED_PROVIDER_EXTERNAL_ENDPOINTS_ENVVAR);
+    
+    const existingCredentialsArray = configuredEndpoints ? JSON.parse(configuredEndpoints)['endpointCredentials'] as EndpointCredentials[] : [];
+    existingCredentialsArray.forEach(cred => {
+        if(!map.has(cred.endpoint)){
+            map.set(cred.endpoint, cred);
         }
+    });
```
AlexVTor, line 185: I notice we aren't warning anymore. Do we want to warn or inform the user that we are not setting creds for an endpoint? If we decide we want to, this looks like a good spot for it

embetten: added it back

AlexVTor (approved):
Looks good! I like the new use of the mapping and cleaning up the junk around the endpoint container.

PR #373, opened 2024-08-29
2202985 Docker V2 task fails to push the container images after the task version updated to 2.243.0

PhilipsonJoseph:
Customer who uses the Podman to run Docker Build would end up in a error is due to difference of layerid creation. DockerCLI used to create with prefix "layerId:sha256" and podman used to create with "layerId:" So to support it we made changes to the code 

#https://github.com/microsoft/azure-pipelines-tasks/issues/20189

### DIFF
--- a/common-npm-packages/docker-common/dockercommandutils.ts
+++ b/common-npm-packages/docker-common/dockercommandutils.ts
@@ -301,13 +301,17 @@ function parseHistoryForLayers(input: string) {
 
 function parseHistoryForV1Name(topHistoryLayer: string): string {
     let v1Name = "";
-    const layerIdString = "layerId:sha256:";
-    const indexOfLayerId = topHistoryLayer.indexOf(layerIdString);
+    let layerIdString = "layerId:sha256:";
+    let indexOfLayerId = topHistoryLayer.indexOf(layerIdString);
+    if (indexOfLayerId < 0) {
+        layerIdString = "layerId:"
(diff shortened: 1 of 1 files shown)


common-npm-packages/docker-common/dockercommandutils.ts
```diff
@@ -301,13 +301,17 @@ function parseHistoryForLayers(input: string) {
...
+    if (indexOfLayerId < 0) {
+        layerIdString = "layerId:"
+        indexOfLayerId = topHistoryLayer.indexOf(layerIdString);
+    }
     if (indexOfLayerId >= 0) {
         v1Name = topHistoryLayer.substring(indexOfLayerId + layerIdString.length);
     }
 
-    return v1Name;
+    return v1Name && v1Name !== "<missing>" ? v1Name : "";
```
z00sts, line 314: Where this "missing" literal comes from?

PhilipsonJoseph: Sometimes layer id will be come as <missing> in that case we need to skip that

z00sts: But why should we handle this case with returning empty string?
`v1Name : ""`

z00sts: I think I get it.

z00sts (approved):

PR #400, opened 2024-10-03
fix(docker): improve retrieval of layer id during inspect

YodaDaCoda:
Should fix https://github.com/microsoft/azure-pipelines-tasks/issues/20506

Accounts for the change from https://github.com/microsoft/azure-pipelines-tasks-common-packages/pull/373 where layerId may or may not include an algorithm.

It appears that when a Dockerfile has a multiline RUN command, `docker history` outputs the `createdBy` property as-is, with newlines. This breaks retrieval of the layerId from the history, as the existing code splits the history by newline and attempts to find the layerId in the first line.

This change uses regex to find the first layerId mentioned in the history, regardless of newlines.


YodaDaCoda:

YodaDaCoda:
or you could just split line-by-line, i'm sure that won't fail in new and exciting ways :unamused:

PR #406, opened 2024-10-24
Changing NuGet cred provider functions when using WIF SC

AlexVTor:
When using a WIF Service connection, we now:
- Verify the feed is external (internal feeds not supported with feedUrl param)
- Don't configure auth for all internal feeds if using an external feed in the same run
- Allow configuration of auth for internal feeds with a WIF Service connection if feed URL is not present

### DIFF
--- a/common-npm-packages/artifacts-common/credentialProviderUtils.ts
+++ b/common-npm-packages/artifacts-common/credentialProviderUtils.ts
@@ -122,7 +122,7 @@ export async function configureCredProvider(protocol: ProtocolType, serviceConne
  * Configure credentials for the given feed using the provided 'Azure Devops' service connection.
  * Only external feeds are supported, will throw if the feed provided is internal.
  */
-export async function configureEntraCredProvider(protocol: ProtocolType, entraWifServiceConnectionName: string, feedUrl: string | undefined) : Promise<boolean> {
+export async function configureEntraCredProvider(protocol: ProtocolType, entraWifServiceConnectionName: string, feedUrl: string | undefined) {
     const connectionData = await getConnectionDataForProtocol(protocol);
     const packagingAccessMappings = getPackagingAccessMappings(connectionData.locationServiceData)
     const allPrefixes: string[] = [...new Set(packagingAccessMappings.map(prefix => prefix.uri))];
@@ -136,7 +136,6 @@ export async function configureEntraCredProvider(protocol: ProtocolType, entraWi
     let token = await getFederatedWorkloadIdentityCredentials(entraWifServiceConnectionName, feedTenant);
     if (token) {
         configureCredProviderForServiceConnectionFeeds([new EntraServiceConnection({uri:feedUrl}, entraWifServiceConnectionName, token)]);
-        return true;
     } else {
         throw new Error(tl.loc("FailedToGetServiceConnectionAuth", entraWifServiceConnectionName)); 
     }
--- a/common-npm-packages/artifacts-common/credentialProviderUtils.ts
+++ b/common-npm-packages/artifacts-common/credentialProviderUtils.ts
@@ -119,18 +119,25 @@ export async function configureCredProvider(protocol: ProtocolType, serviceConne
 }
 
 /**
- * Configure the credential provider to provide credentials for feeds within the pipeline's organization,
- * as well as for a single Entra backed service connection.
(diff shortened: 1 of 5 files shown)


common-npm-packages/artifacts-common/credentialProviderUtils.ts
```diff
@@ -119,36 +119,54 @@ export async function configureCredProvider(protocol: ProtocolType, serviceConne
...
 
 /**
- * Configure the credential provider to provide credentials for feeds within the pipeline's organization,
- * as well as for a single Entra backed service connection.
+ * Configure credentials for the given feed using the provided 'Azure Devops' service connection.
+ * Only external feeds are supported, will throw if the feed provided is internal.
  */
-export async function configureEntraCredProvider(protocol: ProtocolType, feedUrl: string, entraWifServiceConnectionName: string) {
-    await configureCredProviderForSameOrganizationFeeds(protocol);
+export async function configureEntraCredProvider(protocol: ProtocolType, entraWifServiceConnectionName: string, feedUrl: string | undefined) : Promise<boolean> {
```
embetten, line 125: are we certain that we are the only ones using this function and that changing this contract ordering will not break others ?

common-npm-packages/artifacts-common/credentialProviderUtils.ts
```diff
@@ -119,36 +119,54 @@ export async function configureCredProvider(protocol: ProtocolType, serviceConne
...
+    const isFeedInternal = allPrefixes.some(prefix => feedUrl.startsWith(prefix));
+
+    if (isFeedInternal) {
+        throw new Error(tl.loc("Error_InternalFeedsNotSupported"));
+    }
 
     const feedTenant = await getFeedTenantId(feedUrl);
     let token = await getFederatedWorkloadIdentityCredentials(entraWifServiceConnectionName, feedTenant);
     if (token) {
         configureCredProviderForServiceConnectionFeeds([new EntraServiceConnection({uri:feedUrl}, entraWifServiceConnectionName, token)]);
```
embetten, line 138: does this ever return false? or is it only throwing an error in the false case?

common-npm-packages/artifacts-common/credentialProviderUtils.ts
AlexVTor: Yep, I created this function last time around and just confirmed we're the only ones using this

common-npm-packages/artifacts-common/credentialProviderUtils.ts
AlexVTor: Hmm it only throws an error and doesn't return false, only true to show that it was successful. I suppose I can use the throw to exit early and remove the use of the boolean

embetten: I typically prefer booleans to void functions as well, but here I think it might be awkward because the `configureCredProviderForServiceConnectionFeeds` is also void, and there we are not checking if that one is successful. So in the event that getting the token was successful, but `configureCredProviderForServiceConnectionFeeds` silently fails, the returned true here kinda gives off a false sense of success.

common-npm-packages/artifacts-common/module.json
```diff
@@ -16,6 +16,7 @@
         "ErrorInSettingUpSubscription": "Error in setting up subscription",
         "Error_FederatedTokenAquisitionFailed": "Failed to aquire federated token.",
         "Error_GetFeedTenantIdFailed": "Failed to find tenantId for feedUrl. %s",
+        "Error_InternalFeedsNotSupported": "Internal feeds are not supported for use with 'Azure Devops' Service Connections. To use an Azure Devops Service Connection, feedUrl must belong to an external organization.",
```
embetten, line 19: small suggestion to clarify the ado service connection CAN be used internally, just not with feed url

```suggestion
        "Error_InternalFeedsNotSupported": "FeedUrl must belong to an external organization. To authenticate to a feed in your organization with a 'Azure DevOps' service connection type, remove the `feedUrl` input",
```

embetten: slightly odd its in the common repo not the task but if its not possible/easy to move the task repo, please dont let this block

AlexVTor: I agree its a bit odd, Ideally we'd be able to catch specific exceptions from the 
task itself but IIRC, that is more work than its worth. I'd say we can leave it here

PR #411, opened 2024-11-21
Added checksum header to kudu deployment request conditionally

ShilpiRachna1:
In file azure-arm-rest/azure-arm-app-service-kudu.ts, for methods zipDeploy and oneDeploy, 
- based on optional header `addChecksumHeader`, we compute the sha-256 checksum of the zip file attached with the deployment request and add it as a request header `x-ms-artifact-checksum`. 
- This step is skipped if the parameter value is undefined or false. So, there is essentially no change in behaviour if this parameter is not specified or false.
- Any error in computing the checksum is ignored.

This value would be used by the kudu service to validate the integrity of the received file.

Testing done: Validated the changes by deploying update task AzureRmWebAppDeploymentV4


v-schhabra (changes requested):
Hi, could you please build the changes again as I observed "azure-pipelines-tasks-azure-arm-rest" version is not updated in package-lock.json file.

v-schhabra (approved):
LGTM

PR #422, opened 2025-01-10
updating the Get-MsBuildPath function

SimaTian:
to avoid reflection for the new vs studio versions since it causes assembly conflict.

### [Related issue 20734](https://github.com/microsoft/azure-pipelines-tasks/issues/20734)

This portion of the Get-MsBuildPath function script first locates **Microsoft.Build.Utilities.Core.dll** and then uses it to locate the msbuild.exe. Under the current preview version of VS this fails due to an assembly conflict.
Fortunately for us, the reflection usage here is redundant, since the msbuild exe lives either in the same folder as this .dll file or in its direct subfolder.
These paths are fixed due to the way VS installation works - e.g. if we have location of the .dll file, we have the location of both of these versions of msbuild.
```
$Architecture = 'x64'

$Path = "C:\Program Files\Microsoft Visual Studio\2022\Preview\MSBuild\Current\Bin\Microsoft.Build.Utilities.Core.dll"
$msUtilities = [System.Reflection.Assembly]::LoadFrom($Path)

[type]$t = $msUtilities.GetType('Microsoft.Build.Utilities.ToolLocationHelper')
if ($t -ne $null) 
{
    [System.Reflection.MethodInfo] $mi = $t.GetMethod("GetPathToBuildToolsFile",[type[]]@( [string], [string], $msUtilities.GetType("Microsoft.Build.Utilities.DotNetFrameworkArchitecture") ))

$param3 = $mi.GetParameters()[2]
    $archValues = [System.Enum]::GetValues($param3. ParameterType)

[object] $archValue = $null
        if ($Architecture -eq 'x86') {
            $archValue = $archValues.GetValue(1) # DotNetFrameworkArchitecture.Bitness32
        } elseif ($Architecture -eq 'x64') {
            $archValue = $archValues.GetValue(2) # DotNetFrameworkArchitecture.Bitness64
        } else {
            $archValue = $archValues.GetValue(1) # DotNetFrameworkArchitecture.Bitness32
        }
    Write-Host "archValue = $archValue"

$msBuildPath = $mi.Invoke($null, @( 'msbuild.exe', '17.0', $archValue ))
    $msBuildPath
}
```

### DIFF
--- a/common-npm-packages/msbuildhelpers/PathFunctions.ps1
+++ b/common-npm-packages/msbuildhelpers/PathFunctions.ps1
@@ -30,7 +30,6 @@ function Get-MSBuildPath {
         $VersionNumber = [int]($Version.Remove(2))
 
         $specifiedStudio = Get-VisualStudio $VersionNumber
-
         if (($VersionNumber -ge 15 -or !$Version) -and # !$Version indicates "latest"
             ($specifiedStudio = Get-VisualStudio $VersionNumber) -and
             $specifiedStudio.installationPath) {
@@ -39,15 +38,15 @@ function Get-MSBuildPath {
                 if ($VersionNumber -eq 15) {
                     $MsBuildDirectory = "15.0"
                 }
-                Write-Output $MsBuildDirectory
+
                 if ($Architecture -eq 'x86') {
-                    $msBuildPath = Join-Path $specifiedStudio.installationPath MSBuild $MsBuildDirectory Bin MSBuild.exe
+                    $msBuildPath = [System.IO.Path]::Combine($specifiedStudio.installationPath, "MSBuild", $MsBuildDirectory, "Bin\MSBuild.exe");
                     # DotNetFrameworkArchitecture.Bitness32
                 } elseif ($Architecture -eq 'x64') {
-                    $msBuildPath = Join-Path $specifiedStudio.installationPath MSBuild $MsBuildDirectory Bin amd64 MSBuild.exe
+                    $msBuildPath = [System.IO.Path]::Combine($specifiedStudio.installationPath, "MSBuild", $MsBuildDirectory, "Bin\amd64\MSBuild.exe");
                     # DotNetFrameworkArchitecture.Bitness64
                 } else {
-                    $msBuildPath = Join-Path $specifiedStudio.installationPath MSBuild $MsBuildDirectory Bin MSBuild.exe
+                    $msBuildPath = [System.IO.Path]::Combine($specifiedStudio.installationPath, "MSBuild", $MsBuildDirectory, "Bin\MSBuild.exe");
                     # DotNetFrameworkArchitecture.Bitness32
                 }
 
--- a/common-npm-packages/msbuildhelpers/PathFunctions.ps1
+++ b/common-npm-packages/msbuildhelpers/PathFunctions.ps1
@@ -23,36 +23,36 @@ function Get-MSBuildPath {
         [System.Reflection.Assembly]$msUtilities = $null
 
 
-		# We do not need Microsoft.Build.Utilities.Core.dll - if we have located this .dll file, we also have the location of msbuild.exe
-		# for the Bitness32 variant since it resides in the same folder.
-		# and for the Bitness64 variant since it resides in the /amd64 folders.
-		# These paths are fixed due to the way VS is installed.
-		$VersionNumber = [int]($Version.Remove(2))
+        # We do not need Microsoft.Build.Utilities.Core.dll - if we have located this .dll file, we also have the location of msbuild.exe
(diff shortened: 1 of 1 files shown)


SimaTian:
@microsoft-github-policy-service agree [company="{your company}"]

SimaTian:
@microsoft-github-policy-service agree company="Microsoft"

common-npm-packages/msbuildhelpers/PathFunctions.ps1
```diff
@@ -22,16 +22,38 @@ function Get-MSBuildPath {
...
+				}
+
+				if ($msBuildPath -and (Test-Path -LiteralPath $msBuildPath -PathType Leaf)) {
+					Write-Verbose "MSBuild: $msBuildPath"
+					return $msBuildPath
+			}
+		}
+
 
         elseif (($Version -eq "15.0" -or !$Version) -and # !$Version indicates "latest"
```
rainersigwald, line 58: I think you'll have to change this case too, for the "user didn't specify a version" case which will pick up VS17.x

common-npm-packages/msbuildhelpers/PathFunctions.ps1
```diff
@@ -22,17 +22,39 @@ function Get-MSBuildPath {
...
             ($specifiedStudio = Get-VisualStudio $VersionNumber) -and
             $specifiedStudio.installationPath) {
 
-            $msbuildUtilitiesPath = [System.IO.Path]::Combine($specifiedStudio.installationPath, "MSBuild\Current\Bin\Microsoft.Build.Utilities.Core.dll")
-            if (Test-Path -LiteralPath $msbuildUtilitiesPath -PathType Leaf) {
-                Write-Verbose "Loading $msbuildUtilitiesPath"
-                $msUtilities = [System.Reflection.Assembly]::LoadFrom($msbuildUtilitiesPath)
+                [object]$archValue = $null
+                if ($Architecture -eq 'x86') {
+                    $msBuildPath = [System.IO.Path]::Combine($specifiedStudio.installationPath, "MSBuild\Current\Bin\msbuild.exe")
```
rainersigwald, line 41: This matches what was here before but I'd expect something more like
```suggestion
                    $msBuildPath = Join-Path $specifiedStudio.installationPath MSBuild Current Bin MSBuild.exe
```

rainersigwald (commented):
I strongly agree with the direction here.

common-npm-packages/msbuildhelpers/PathFunctions.ps1
SimaTian: Isn't there the `!$Version branch` in the above portion of the script as well?
That being said, I can do the update here since the only difference I could see is in the `MSBuild\15.0\Bin\` vs `MSBuild\Current\Bin\` portion of the path so it should work in the same way.

 I was trying to keep my changes to a minimum to be on the safe side.

common-npm-packages/msbuildhelpers/PathFunctions.ps1
```diff
@@ -22,17 +22,39 @@ function Get-MSBuildPath {
...
             ($specifiedStudio = Get-VisualStudio $VersionNumber) -and
             $specifiedStudio.installationPath) {
 
-            $msbuildUtilitiesPath = [System.IO.Path]::Combine($specifiedStudio.installationPath, "MSBuild\Current\Bin\Microsoft.Build.Utilities.Core.dll")
-            if (Test-Path -LiteralPath $msbuildUtilitiesPath -PathType Leaf) {
-                Write-Verbose "Loading $msbuildUtilitiesPath"
-                $msUtilities = [System.Reflection.Assembly]::LoadFrom($msbuildUtilitiesPath)
+                [object]$archValue = $null
+                if ($Architecture -eq 'x86') {
+                    $msBuildPath = Join-Path $specifiedStudio.installationPath MSBuild Current Bin MSBuild.exe
```
SimaTian, line 41: This works with powershell 7.4+
On my machine it breaks with Windows Powershell since I probably have some older version.
Is that something we should be concerned about?

rainersigwald: Ok that's a very convincing reason!

(As a rule Windows PowerShell 5.1 is the only "real" one; the new cross-platform ones are much better but since they don't come with Windows can't be relied upon in the same way.)

SimaTian: Ok, I will return to the old way of doing the string manipulation.

praval-microsoft:
@SimaTian we will be adding this change under a feature flag. https://github.com/microsoft/azure-pipelines-tasks-common-packages/pull/424

PR #424, opened 2025-01-16
Adding Get-MSBuildPathModified with Telemetry under Feature flags

praval-microsoft:
to avoid reflection for the new vs studio versions since it causes assembly conflict.

### [Related issue 20734](https://github.com/microsoft/azure-pipelines-tasks/issues/20734)

This portion of the Get-MsBuildPath function script first locates **Microsoft.Build.Utilities.Core.dll** and then uses it to locate the msbuild.exe. Under the current preview version of VS this fails due to an assembly conflict.
Fortunately for us, the reflection usage here is redundant, since the msbuild exe lives either in the same folder as this .dll file or in its direct subfolder.
These paths are fixed due to the way VS installation works - e.g. if we have location of the .dll file, we have the location of both of these versions of msbuild.
```
$Architecture = 'x64'

$Path = "C:\Program Files\Microsoft Visual Studio\2022\Preview\MSBuild\Current\Bin\Microsoft.Build.Utilities.Core.dll"
$msUtilities = [System.Reflection.Assembly]::LoadFrom($Path)

[type]$t = $msUtilities.GetType('Microsoft.Build.Utilities.ToolLocationHelper')
if ($t -ne $null) 
{
    [System.Reflection.MethodInfo] $mi = $t.GetMethod("GetPathToBuildToolsFile",[type[]]@( [string], [string], $msUtilities.GetType("Microsoft.Build.Utilities.DotNetFrameworkArchitecture") ))

$param3 = $mi.GetParameters()[2]
    $archValues = [System.Enum]::GetValues($param3. ParameterType)

[object] $archValue = $null
        if ($Architecture -eq 'x86') {
            $archValue = $archValues.GetValue(1) # DotNetFrameworkArchitecture.Bitness32
        } elseif ($Architecture -eq 'x64') {
            $archValue = $archValues.GetValue(2) # DotNetFrameworkArchitecture.Bitness64
        } else {
            $archValue = $archValues.GetValue(1) # DotNetFrameworkArchitecture.Bitness32
        }
    Write-Host "archValue = $archValue"

$msBuildPath = $mi.Invoke($null, @( 'msbuild.exe', '17.0', $archValue ))
    $msBuildPath
}
```

### Testing
- Tested the Feature Flag logic by deploying the MSBuildTask with Updated MSBuildHelpers package in a personal test organisation. 
- Verified for the Visual Studio 15,16 and 17 for both Architectures x64 and x86.

### DIFF
--- a/common-npm-packages/msbuildhelpers/PathFunctions.ps1
+++ b/common-npm-packages/msbuildhelpers/PathFunctions.ps1
@@ -263,6 +263,7 @@ function Select-MSBuildPath {
 
     $selectMSBuildPathTelemetry = [PSCustomObject]@{
         PreferredVersion = $PreferredVersion
+        LookedUpVersion = ""
         Architecture = $Architecture
         PathFromGetMSBuildPathFunction = ""
         PathFromGetMSBuildPathFunctionV2 = ""
@@ -302,8 +303,10 @@ function Select-MSBuildPath {
             if (($path = Get-MSBuildPath -Version $PreferredVersion -Architecture $Architecture)) {
                 if($featureFlags.enableTelemetry) {
                     try {
-                        $pathFromGetMSBuildPathV2 = Get-MSBuildPathV2 -Version $PreferredVersion -Architecture $Architecture
+                        $selectMSBuildPathTelemetry.LookedUpVersion = $PreferredVersion
                         $selectMSBuildPathTelemetry.PathFromGetMSBuildPathFunction = $path
+
+                        $pathFromGetMSBuildPathV2 = Get-MSBuildPathV2 -Version $PreferredVersion -Architecture $Architecture
                         $selectMSBuildPathTelemetry.PathFromGetMSBuildPathFunctionV2 = $pathFromGetMSBuildPathV2
                         $selectMSBuildPathTelemetry.PathMatches = ($path -eq $pathFromGetMSBuildPathV2)
                     } catch {
@@ -327,8 +330,10 @@ function Select-MSBuildPath {
             if (($path = Get-MSBuildPath -Version $version -Architecture $Architecture)) {
                 if($featureFlags.enableTelemetry) {
                     try {
-                        $pathFromGetMSBuildPathV2 = Get-MSBuildPathV2 -Version $version -Architecture $Architecture
+                        $selectMSBuildPathTelemetry.LookedUpVersion = $version
                         $selectMSBuildPathTelemetry.PathFromGetMSBuildPathFunction = $path
+
+                        $pathFromGetMSBuildPathV2 = Get-MSBuildPathV2 -Version $version -Architecture $Architecture
                         $selectMSBuildPathTelemetry.PathFromGetMSBuildPathFunctionV2 = $pathFromGetMSBuildPathV2
                         $selectMSBuildPathTelemetry.PathMatches = ($path -eq $pathFromGetMSBuildPathV2)
                     } catch {
@@ -338,12 +343,10 @@ function Select-MSBuildPath {
                     EmitTelemetry -TelemetryPayload $selectMSBuildPathTelemetry -TaskName "MSBuildHelpers"
                     
                     if($featureFlags.useGetMSBuildPathV2) {
-                        
                         # Warn falling back.
                         if ($specificVersion) {
                             Write-Warning (Get-VstsLocString -Key 'MSB_UnableToFindMSBuildVersion0Architecture1FallbackVersion2' -ArgumentList $PreferredVersion, $Architecture, $version)
                         }
-
                         return $pathFromGetMSBuildPathV2
                     }
                 }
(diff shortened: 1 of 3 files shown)


common-npm-packages/msbuildhelpers/PathFunctions.ps1
```diff
@@ -292,11 +499,33 @@ function Select-MSBuildPath {
         # Look for the latest version of MSBuild.
         foreach ($version in $versions) {
             if (($path = Get-MSBuildPath -Version $version -Architecture $Architecture)) {
```
tarunramsinghani, line 501: The telemetry and v2 function will not be executed if value from GetMsBuildPath is null for some reason, Should we move the whole telemetry and return v2 outside of this if condition ?

common-npm-packages/msbuildhelpers/PathFunctions.ps1
```diff
@@ -252,6 +426,22 @@ function Select-MSBuildPath {
         [string]$Architecture)
 
     Trace-VstsEnteringInvocation $MyInvocation
+
+    $featureFlags = @{
+        enableTelemetry  = [System.Convert]::ToBoolean($env:MSBUILDHELPERS_PATHFUNCTIONS_ENABLE_TELEMETRY)
```
merlynomsft, line 431: bug bug.  It looks like this will raise an error if the env variable is an empty string

common-npm-packages/msbuildhelpers/PathFunctions.ps1
```diff
@@ -160,6 +162,177 @@ function Get-MSBuildPath {
...
+# Under the current preview version of VS 17.3.2 this fails due to an assembly conflict.
+# the reflection usage here is redundant, since the msbuild exe lives either in the same folder as this .dll file or in its direct subfolder.
+# [TODO] : Once we have enough telemetry records to be confident of the correctness of this method, replace Get-MSBuildPath with Get-MSBuildPathV2 and remove feature flags.
+function Get-MSBuildPathV2 {
+    [CmdletBinding()]
+    param(
+        [string]$Version,
+        [string]$Architecture)
+
+    $VersionNumber = [int]($Version.Remove(2))
```
merlynomsft, line 176: remove can throw if 2 is longer than the string

common-npm-packages/msbuildhelpers/PathFunctions.ps1
praval-microsoft: I have actually tested it. It returns false in case of null and empty string
image

common-npm-packages/msbuildhelpers/PathFunctions.ps1
```diff
@@ -281,22 +471,69 @@ function Select-MSBuildPath {
...
+
+                EmitTelemetry -TelemetryPayload $selectMSBuildPathTelemetry -TaskName "MSBuildHelpers"
+
+                if($featureFlags.useGetMSBuildPathV2 -and $pathFromGetMSBuildPathV2) {
+                    Write-Debug "Returning path from GetMSBuildPathV2"
+                    return $pathFromGetMSBuildPathV2
+                }
+            }
+            
+            if($path) {
```
merlynomsft, line 496: I'd suggest avoiding changing this line.  It appears the same, but the semantics in powershell can be tricky.  Better to not change it as it's not guarded by the feature flag

Instead we can just a different variable for $path in the enableTelemetry code path

common-npm-packages/msbuildhelpers/PathFunctions.ps1
```diff
@@ -281,22 +471,69 @@ function Select-MSBuildPath {
...
-            if (($path = Get-MSBuildPath -Version $PreferredVersion -Architecture $Architecture)) {
+            $path = Get-MSBuildPath -Version $PreferredVersion -Architecture $Architecture
+            $pathFromGetMSBuildPathV2 = $null
+
+            if($featureFlags.enableTelemetry) {
+                try {
+                    $selectMSBuildPathTelemetry.LookedUpVersion = $PreferredVersion
+                    $selectMSBuildPathTelemetry.PathFromGetMSBuildPathFunction = $path
+                    $pathFromGetMSBuildPathV2 = Get-MSBuildPathV2 -Version $PreferredVersion -Architecture $Architecture
+                    $selectMSBuildPathTelemetry.PathFromGetMSBuildPathFunctionV2 = $pathFromGetMSBuildPathV2
```
merlynomsft, line 482: (shared offline comment)

common-npm-packages/msbuildhelpers/PathFunctions.ps1
tarunramsinghani: Looks good

common-npm-packages/msbuildhelpers/PathFunctions.ps1
praval-microsoft: Updated the Feature flag input logic

common-npm-packages/msbuildhelpers/PathFunctions.ps1
praval-microsoft: Updated

common-npm-packages/msbuildhelpers/PathFunctions.ps1
praval-microsoft: Removed

merlynomsft (approved):

PR #427, opened 2025-01-24
Added functions to the NuGetMockHelper for mocking the os and fs modules.

amp-powell:
Added functions to the NuGetMockHelper for mocking the os and fs modules.


cobya:
Replaced by #432

PR #432, opened 2025-01-30
Add NuGet os and fs module mocking functions

cobya:
Replacement of #427 

Adding mockable 'os' and 'fs' modules to support the rollout of NuGetCommandV2 updates for the Ubuntu 24.04 versions without mono.


magleaso (approved):
LGTM

PR #434, opened 2025-02-20
Add execAsync wrapper for NuGetToolRunner2

cobya:
Add a replacement wrapped for the deprecated `exec` method in the `NuGetToolRunner2` class.


embetten (approved):
LGTM

PR #448, opened 2025-04-17
modified docker label image code as per suggestions

v-dko:
Fix Label image.base.digest when base image is specified in Dockerfile FROM with digest instead of tag

PR from public: https://github.com/microsoft/azure-pipelines-tasks-common-packages/pull/314


common-npm-packages/docker-common/containerimageutils.ts
```diff
@@ -304,7 +301,7 @@ export function getBaseImageDigestDockerFile(dockerFileContentPath: string): str
             }
         }
 
-        return null;
+        return baseImageData[0];
```
brbayes-msft, line 304: This would end up returning the container image name if no digest is present.  Based on the code comment at the top of this method, isn't the desired behavior to return null if no digest is present?  This would also seem to align with the call from pipelineutils.ts, which falls back to the old logic if no digest is found in the FROM line.

common-npm-packages/docker-common/containerimageutils.ts
```diff
@@ -253,60 +253,66 @@ export function getImageIdFromBuildOutput(output: string): string {
 
     return "";
 }
-export function getBaseImageDigestDockerFile(dockerFileContentPath: string): string {
+export function getBaseImageDetialsFromDockerFIle(dockerFileContentPath: string, connection?: ContainerConnection): any {
```
brbayes-msft, line 256: nit: I might suggest a stronger return type than any to maintain strong typings.

brbayes-msft (approved):

common-npm-packages/docker-common/containerimageutils.ts
v-dko: added return type instead of any

manolerazvan (approved):

PR #458, opened 2025-05-23
Add ability to build one common package

ivanduplenskikh:
**Description**: This PR adds ability to build tasks separatelly

### DIFF
--- a/make.js
+++ b/make.js
@@ -27,21 +27,32 @@ const printLabel = (name) => {
     console.log('----------------------------------');
 }
 
+const installBuildScriptsDependencies = () => {
+    console.log('Building Tests');
+    util.cd('common-npm-packages/build-scripts');
+    util.run('npm install');
+    util.cd(__dirname);
+}
+
 const buildPsTestHelpers = () => {
     console.log('Building Tests');
     util.cd('Tests');
     util.run('npm install');
     util.run(path.join('node_modules', '.bin', 'tsc'));
-    util.cd('..');
+    util.cd(__dirname);
 }
+
+installBuildScriptsDependencies();
 
 if (options.build) {
     buildPsTestHelpers();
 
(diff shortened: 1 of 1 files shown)


make.js
```diff
@@ -27,21 +27,32 @@ const printLabel = (name) => {
     console.log('----------------------------------');
 }
 
+const installBuildScriptsDependencies = () => {
+    console.log('Building Tests');
```
tarunramsinghani, line 31: Log message seems wrong. This should be "Installing dependencies for BuildScripts" instead ?

make.js
```diff
@@ -27,21 +27,32 @@ const printLabel = (name) => {
...
+
 if (options.build) {
     buildPsTestHelpers();
 
     console.log('\nBuilding shared npm packages');
     util.cd('common-npm-packages');
+
     fs.readdirSync('./', { encoding: 'utf-8' }).forEach(child => {
         if (fs.statSync(child).isDirectory() && !ignoredFolders.includes(child)) {
+            if (options.task && child !== options.task) return;
```
tarunramsinghani, line 55: Can we rename task variable to "packageName" and add it to predefinedFlags above.

make.js
```diff
@@ -86,7 +98,7 @@ if (options.test) {
                     console.log('No tests found for the package');
                 }
             } else {
-                throw new Error('Package has not been built');
+                throw new Error(`Package "${buildPath}" has not been built`);
```
tarunramsinghani, line 101: NIT : we can use "Options.task" instead of "${buildPath}" to be more clear.

tarunramsinghani:
Please add the details on how to use this new functionality in the PR comments here and we should update the documentation as well in the ADO wiki.

tarunramsinghani (approved):

PR #464, opened 2025-06-10
Add an exponential backoff mechanism for obtaining tokens

ivanduplenskikh:
**Description**: initOIDCToken now waits for waitedTime * retryCount in milliseconds to prevent inreliable failed responses

**Related WI**: [AB#2275633](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2275633)

### DIFF
--- a/common-npm-packages/azure-arm-rest/azCliUtility.ts
+++ b/common-npm-packages/azure-arm-rest/azCliUtility.ts
@@ -204,7 +204,7 @@ export async function initOIDCToken2(
         }
 
         if (retryCount >= MAX_CREATE_OIDC_TOKEN_RETRIES) {
-            return Promise.reject(tl.loc('CouldNotFetchAccessTokenForAADRetryLimitExceeded'));
+            throw Error(tl.loc('CouldNotFetchAccessTokenForAADRetryLimitExceeded'));
         }
 
         retryCount += 1;
--- a/common-npm-packages/azure-arm-rest/azCliUtility.ts
+++ b/common-npm-packages/azure-arm-rest/azCliUtility.ts
@@ -148,7 +148,7 @@ export async function initOIDCToken(
 
         if (token?.oidcToken !== null && token?.oidcToken !== undefined) {
             tl.debug('Got OIDC token');
-            return Promise.resolve(token.oidcToken);
+            return token.oidcToken;
         }
 
         // If the token is null, it means the OIDC token could not be fetched
--- a/common-npm-packages/azure-arm-rest/azCliUtility.ts
+++ b/common-npm-packages/azure-arm-rest/azCliUtility.ts
@@ -131,7 +131,16 @@ export async function getFederatedToken(connectedServiceName: string): Promise<s
     return oidc_token;
 }
 
-export async function initOIDCToken(connection: WebApi, projectId: string, hub: string, planId: string, jobId: string, serviceConnectionId: string, retryCount: number = 0, timeToWait: number = 2000): Promise<string> {
+export async function initOIDCToken(
+    connection: WebApi,
+    projectId: string,
+    hub: string,
+    planId: string,
+    jobId: string,
+    serviceConnectionId: string,
+    retryCount: number = 0,
+    timeToWait: number = 2000
+): Promise<string> {
     const taskApi = await connection.getTaskApi();
 
     try {
--- a/common-npm-packages/azure-arm-rest/azCliUtility.ts
+++ b/common-npm-packages/azure-arm-rest/azCliUtility.ts
@@ -169,14 +169,11 @@ export async function initOIDCToken(
         }
 
         retryCount += 1;
-
         tl.debug(`Retrying OIDC token fetch. Retries left: ${MAX_CREATE_OIDC_TOKEN_RETRIES - retryCount}`);
 
-        return new Promise((resolve) => {
-            setTimeout(() => {
-                resolve(initOIDCToken(connection, projectId, hub, planId, jobId, serviceConnectionId, retryCount, timeToWait));
-            }, Math.min(timeToWait * retryCount, MAX_CREATE_OIDC_TOKEN_BACKOFF_TIMEOUT));
-        });
+        // Wait for a backoff time before retrying
+        await new Promise(resolve => setTimeout(resolve, Math.min(timeToWait * retryCount, MAX_CREATE_OIDC_TOKEN_BACKOFF_TIMEOUT)));
+        return initOIDCToken(connection, projectId, hub, planId, jobId, serviceConnectionId, retryCount, timeToWait);
     }
 }
 
--- a/common-npm-packages/azure-arm-rest/azCliUtility.ts
+++ b/common-npm-packages/azure-arm-rest/azCliUtility.ts
@@ -149,7 +149,7 @@ export async function initOIDCToken2(
     try {
         const token = await taskApi.createOidcToken({}, projectId, hub, planId, jobId, serviceConnectionId);
 
-        if (token?.oidcToken) {
+        if (!token?.oidcToken) {
             tl.debug('Got OIDC token');
             return token.oidcToken;
         }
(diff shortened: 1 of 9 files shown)


common-npm-packages/azure-arm-rest/azCliUtility.ts
```diff
@@ -130,37 +128,42 @@ export async function getFederatedToken(connectedServiceName: string): Promise<s
...
-    );
+        return Promise.reject(tl.loc('CouldNotFetchAccessTokenforAAD') + " " + error);
+    }
+
+    if (token != null) {
+        tl.debug('Got OIDC token');
+        return Promise.resolve(token.oidcToken);
+    }
+
+    if (token.oidcToken === null && retryCount >= 3) {
```
aaudzei, line 156: I understand that this is equivalent block from previous version of `function initOIDCToken` but I think this condition will never be true due to the `if (token != null) {` block above.

common-npm-packages/azure-arm-rest/azCliUtility.ts
```diff
@@ -130,37 +128,42 @@ export async function getFederatedToken(connectedServiceName: string): Promise<s
...
+        return Promise.reject(tl.loc('CouldNotFetchAccessTokenforAAD'));
+    }
+
+    retryCount += 1;
 
-    return deferred.promise;
+    return new Promise((resolve) => {
+        setTimeout(() => {
+            resolve(initOIDCToken(connection, projectId, hub, planId, jobId, serviceConnectionId, retryCount, timeToWait));
+        }, timeToWait * retryCount);
```
aaudzei, line 165: It is a good practice to limit maximum delay in case of exponential backoffs because they have a tendency to rapidly increase to the level that makes no sense.

Here is an example from another MS library: https://github.com/microsoft/azure-pipelines-tasks-common-packages/blob/d4d4458180f1feb4de6a4216061ba224d34fc56f/common-npm-packages/azure-arm-rest/azure-arm-common.ts#L469

common-npm-packages/azure-arm-rest/azCliUtility.ts
```diff
@@ -130,37 +128,42 @@ export async function getFederatedToken(connectedServiceName: string): Promise<s
...
-                        else {
-                            deferred.reject(tl.loc('CouldNotFetchAccessTokenforAAD'));
-                        }
-                    }
-                },
-                (error) => {
-                    deferred.reject(tl.loc('CouldNotFetchAccessTokenforAAD') + " " + error);
-                }
-            );
+export async function initOIDCToken(connection: WebApi, projectId: string, hub: string, planId: string, jobId: string, serviceConnectionId: string, retryCount: number = 0, timeToWait: number = 2000): Promise<string> {
```
aaudzei, line 131: We briefly discussed that `retryCount` has different semantical meaning in `common-npm-packages/azure-arm-rest/azure-arm-common.ts` so the unification requires more care.

With this implementation and call originating from (here)[https://github.com/microsoft/azure-pipelines-tasks-common-packages/blob/d4d4458180f1feb4de6a4216061ba224d34fc56f/common-npm-packages/azure-arm-rest/azure-arm-common.ts#L469] there will no retries at all as far as I can tell.

aaudzei (changes requested):
the `retryCount` unification is a breaking change so blocking the PR in its current state

common-npm-packages/azure-arm-rest/azure-arm-common.ts
```diff
@@ -214,28 +215,7 @@ export class ApplicationTokenCredentials {
...
-                const response = await api.createOidcToken({}, projectId, hub, planId, jobId, serviceConnectionId);
-                if (response && response.oidcToken) {
-                    tl.debug('Got OIDC token');
-                    return response.oidcToken;
-                }
-            } catch (e: any) {
-                error = e;
-            }
-            await new Promise(r => setTimeout(r, timeToWait));
-            tl.debug(`Retrying OIDC token fetch. Retries left: ${i}`);
```
aaudzei, line 230: I think new code is missing equivalent log line and I think it can useful for troubleshooting

common-npm-packages/azure-arm-rest/azCliUtility.ts
```diff
@@ -130,37 +128,42 @@ export async function getFederatedToken(connectedServiceName: string): Promise<s
...
+    let token: TaskHubOidcToken;
+
+    try {
+        token = await taskApi.createOidcToken({}, projectId, hub, planId, jobId, serviceConnectionId)
+    } catch (error) {
+        // Handle AggregateError if available, since the package uses Node10 types.
+        // Otherwise handle generic error
+        if (error.name === 'AggregateError' && error['errors'] !== undefined) {
+            for (const err of error.errors) {
+                tl.error(err);
```
aaudzei, line 143: It may be hard to understand from the logs where the error is coming from
```suggestion
                tl.error("Error while trying to get OIDC token: " + err);
```

ivanduplenskikh: Added to either condition when the error is an instance of AggregateError and the other type

common-npm-packages/azure-arm-rest/azCliUtility.ts
```diff
@@ -121,46 +124,50 @@ export async function getFederatedToken(connectedServiceName: string): Promise<s
...
+            for (const err of error.errors) {
+                tl.error(`Error while trying to get OIDC token: ${err}`);
+            }
+        } else {
+            tl.error(`Error while trying to get OIDC token: ${error}`);
         }
-    );
 
-    return deferred.promise;
+        if (retryCount >= MAX_CREATE_OIDC_TOKEN_RETRIES) {
```
aaudzei, line 159: Previous version of the function would have tried 3 times until `if (retryCount < 3) {`
But new version will only try twice. That's because `retryCount += 1;` happens before checking for max here.
Please double check if my math is correct and if so fix the condition but also `tl.debug(`Retrying OIDC token fetch ...` below

common-npm-packages/azure-arm-rest/azure-arm-common.ts
```diff
@@ -214,28 +215,7 @@ export class ApplicationTokenCredentials {
...
-        }
-
-        let message = tl.loc('CouldNotFetchAccessTokenforAAD');
-        if (error) {
-            message += " " + error;
-        }
-
-        return Promise.reject(message);
+        return await azCliUtility.initOIDCToken(connection, projectId, hub, planId, jobId, serviceConnectionId, retryCount, timeToWait);
```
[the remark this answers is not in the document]
aaudzei: It is quite possible but I don't think it has to be part of this PR.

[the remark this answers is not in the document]
ivanduplenskikh: Fixed

common-npm-packages/azure-arm-rest/azCliUtility.ts
```diff
@@ -121,46 +124,60 @@ export async function getFederatedToken(connectedServiceName: string): Promise<s
...
+    timeToWait: number = 2000
+): Promise<string> {
+    const taskApi = await connection.getTaskApi();
+
+    try {
+        const token = await taskApi.createOidcToken({}, projectId, hub, planId, jobId, serviceConnectionId);
+
+        if (token?.oidcToken !== null && token?.oidcToken !== undefined) {
+            tl.debug('Got OIDC token');
+            return Promise.resolve(token.oidcToken);
```
onetocny, line 151: Is `Promise.resolve` necessary here? Can't you just return `token.oidcToken`?

common-npm-packages/azure-arm-rest/azCliUtility.ts
```diff
@@ -121,46 +124,60 @@ export async function getFederatedToken(connectedServiceName: string): Promise<s
...
+        if (error.name === 'AggregateError' && error['errors'] !== undefined) {
+            for (const err of error.errors) {
+                tl.error(`Error while trying to get OIDC token: ${err}`);
+            }
+        } else {
+            tl.error(`Error while trying to get OIDC token: ${error}`);
+        }
+
+        if (retryCount >= MAX_CREATE_OIDC_TOKEN_RETRIES) {
+            return Promise.reject(tl.loc('CouldNotFetchAccessTokenForAADRetryLimitExceeded'));
```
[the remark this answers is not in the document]
onetocny: As you have rewritten the code into async/await model it would appropriate to convert the `reject` to `throw`. Previous implementation is using `reject` handler passed to `Promise` constructor. However, here you are using `Promise.reject` static method which is not the same. Feel free to wont fix.

onetocny (approved):

PR #630, opened 2026-06-18
Sanitize Docker output to prevent ##vso[] logging-command injection

onetocny:
### **Description**

**Security fix: Logging-command injection via Docker task output (MSRC Case 122404 / VULN-195611)**

AB#2378653

The Azure Pipelines Docker tasks (`Docker@2`, `ContainerBuild@0`) stream child process stdout/stderr directly to the agent via `ToolRunner.exec()`. The agent scans everything written to process.stdout for `##vso[]` prefixes and interprets them as logging commands. An attacker with queue-builds permission can craft a Dockerfile that emits `##vso[task.prependpath]` (or `task.setvariable`, etc.) via `RUN echo`, hijacking the PATH or injecting variables for subsequent pipeline steps — achieving arbitrary code execution.

**Fix:** Wrap `outStream`/`errStream` passed to `ToolRunner.exec()` with a sanitizing `Writable` stream that replaces `##vso[` with `#vso[` (case-insensitive) before the agent sees it. The raw output is still available via the `stdout`/`stderr` event listeners for internal parsing (image ID extraction, digest capture, etc.).

**Impact:** All 5 public command functions (`build`, `command`, `push`, `start`, `stop`) now use sanitized exec options. Existing behavior is preserved — only the agent-visible output is neutered.

---

### **Package Name**

`azure-pipelines-tasks-docker-common`

---

**Low**

- The fix is additive — it wraps the output stream without changing any command logic or argument handling.
- The sanitization only affects what the agent sees on stdout/stderr; internal output callbacks still receive raw data for parsing (image IDs, digests).
- The replacement (`##vso[` → `#vso[`) is safe — legitimate pipeline logging commands are never emitted by Docker itself.
- Backward compatible: no API changes, no new dependencies beyond `mocha` (devDependency for tests).

---

- [x] Manual tests performed

14 new unit tests covering all 5 public functions:
- Verifies `outStream`/`errStream` options are provided and are NOT raw `process.stdout`/`process.stderr`
- Verifies `##vso[task.prependpath]` is sanitized in stdout
- Verifies `##vso[task.setvariable]` is sanitized in stderr (BuildKit scenario)
- Verifies raw output callback still receives unsanitized data (for image ID extraction)
- Verifies clean Docker output passes through unmodified
- Verifies case-insensitive variants (`##VSO[`) are caught
- Verifies multiple `##vso[` commands in a single chunk are all sanitized

---

End-to-end validation in `onetocny-corp` Azure DevOps organization:

1. **Reproduced the vulnerability** — Created a pipeline with a Dockerfile containing `RUN echo "##vso[task.prependpath]/tmp/pwned"`. Confirmed `/tmp/pwned` appeared in PATH of subsequent steps.

2. **Validated Docker@2 fix** — Built the task with the patched `docker-common` package (via `npm pack` + local `.tgz` reference per wiki process), uploaded to org via `tfx`, reran pipeline → `PATH is clean — injection was blocked.` ✅

3. **Validated ContainerBuild@0 fix** — Same process for the only other affected task → `PATH is clean - injection was blocked.` ✅

Pipeline runs: Docker2-Injection-Repro (run 32), ContainerBuild0-Injection-Repro (run 33) in `onetocny-corp/playground`.

---

No — this is an internal implementation change with no public API surface modifications.

---

### **Dependencies**

- `mocha` `^11.7.5` added as a **devDependency** (for unit test execution only, not shipped with the package)

---

### **Checklist**

- [x] Package version was bumped — see [versioning guide](https://semver.org/)
- [x] CLA signed (if applicable) — see [Contributor License Agreement](https://cla.opensource.microsoft.com)

### DIFF
--- a/common-npm-packages/docker-common/Tests/L0.ts
+++ b/common-npm-packages/docker-common/Tests/L0.ts
@@ -1,0 +1,5 @@
+import { runDockerCommandSanitizationTests } from './dockerOutputSanitizationTests';
+
+describe('docker-common suite', () => {
+    describe('Docker command output sanitization', runDockerCommandSanitizationTests);
+});
--- a/common-npm-packages/docker-common/Tests/dockerOutputSanitizationTests.ts
+++ b/common-npm-packages/docker-common/Tests/dockerOutputSanitizationTests.ts
@@ -1,0 +1,309 @@
+import assert = require("assert");
+import { EventEmitter } from "events";
+import { Writable } from "stream";
+
+// Set environment variables required by azure-pipelines-task-lib before importing dockerCommandUtils
+process.env['INPUT_BUILDCONTEXT'] = '/tmp/build';
+process.env['SYSTEM_DEFAULTWORKINGDIRECTORY'] = '/tmp/work';
+
+import * as dockerCommandUtils from "../dockercommandutils";
+
+/**
+ * Minimal mock for ToolRunner that simulates Docker command execution.
+ * When exec() is called, it emits stdout/stderr data and writes to the
+ * provided outStream/errStream options, mimicking real ToolRunner behavior.
+ */
+class MockToolRunner extends EventEmitter {
+    public simulatedStdout: string = "";
+    public simulatedStderr: string = "";
+
+    arg(_val: string | string[]): void { }
+    line(_val: string): void { }
+
+    exec(options?: any): Promise<void> {
+        if (this.simulatedStdout) {
+            this.emit("stdout", this.simulatedStdout);
+        }
+        if (this.simulatedStderr) {
+            this.emit("stderr", this.simulatedStderr);
+        }
+
+        return new Promise<void>((resolve) => {
+            const writes: Promise<void>[] = [];
+
+            if (this.simulatedStdout && options && options.outStream) {
+                writes.push(new Promise<void>((res) => {
+                    options.outStream.write(this.simulatedStdout, 'utf8', () => res());
+                }));
+            }
+            if (this.simulatedStderr && options && options.errStream) {
+                writes.push(new Promise<void>((res) => {
+                    options.errStream.write(this.simulatedStderr, 'utf8', () => res());
+                }));
+            }
+
+            Promise.all(writes).then(() => resolve());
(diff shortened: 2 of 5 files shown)


wawanawna (changes requested):
Ondrej please address these 2:

 1. The sanitizer is evadable (chunk‑split). ToolRunner writes raw pipe Buffer chunks to the wrapper ( toolrunner.ts  →   outStream.write(data) ), and  sanitizeDockerOutput  runs per‑chunk with no cross‑chunk state. If  ##vso[  is split across two  chunks it passes through —  process.stdout  reassembles it and the agent acts on it. Reproduced: writing  ##vs  then   o[task.prependpath]/tmp/pwned  yields  ##vso[…]  downstream. An attacker can pad  RUN echo  output to land the token on a pipe  read boundary (moderate effort, but a string‑replace that breaks when the token is split isn't robust enough for a security  control).
 Fix: make the sanitizer stateful — keep a 5‑char carry‑over tail between writes, prepend before scanning, flush on  final ;  decode with  StringDecoder('utf8')  (also fixes multibyte mojibake from  chunk.toString()  on split UTF‑8).

 2.  getHistory  is unsanitized and reachable. Only  build/command/push/start/stop  were wrapped.  getHistory   ( dockercommandutils.ts:344 ) runs  docker history --no-trunc , whose  createdBy  column replays the Dockerfile  RUN  text  verbatim to unsanitized stdout, and it runs on every push/buildAndPush via  dockerpush.ts:114 publishToImageMetadataStore → :139  getHistory . The agent matches  ##vso[  mid‑line ( Command.cs   IndexOf , no starts‑with gate), so it fires. (The injected  value is polluted by the trailing  "; layerId:…  from the  --format , which makes clean exploitation harder but doesn't stop the  command.)
 Fix: route this path through the sanitized stream too, and audit the other  execCommand  callers —  getImageRootfsLayers  is  fine (digests only).

 Also: add a chunk‑split unit test — all 14 tests emit the token in a single chunk, so they don't cover the bypass above

wawanawna (approved):

PR #543, opened 2025-12-23
Fix package source mapping keys in temp config file For packing-common

deekshithamerugu-iress:
### **Description**
NuGet Package Source Mapping enforces that a package's declared source in <packageSourceMapping> must match a corresponding source name under <packageSources>.
The Azure Pipelines internal NuGet helper (NuGetConfigHelper2.ts) currently prefixes internal feed names with feed- when writing the temporary NuGet.config used during restore,
However, the helper does not update the <packageSourceMapping> section accordingly, leaving the mapping with the original key:

NuGet interprets this mismatch as a violation of Package Source Mapping rules and fails with:

> NU1100: PackageSourceMapping is enabled, the following source(s) were not considered: feed-canarytest

📌 [Bug 2340656](https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2340656): DotNetCoreCLI@2 breaks package source mappings for internal feeds when using a nuget.config file

Keeps existing feed prefixing untouched (feed-<name> logic stays exactly as before).
Tracks original → prefixed feed names when internal sources are rewritten.
After the temp NuGet.config is generated, the fix:
Loads the XML,
Locates <packageSourceMapping>,
Updates mapping keys to match their prefixed feed names, e.g.: canarytest → feed-canarytest
Writes back the corrected temporary config.

---

### **Package Name**
packaging-common

---

### **Risk Assessment** (Low / Medium / High)  
Low - Only modifies temporary NuGet.config used during restore and doesnot modify prefixing logic or authentication logic.

---

### **Unit Tests Added or Updated**   
- [ ] Unit tests added or updated
- [x] Manual tests performed

---

### **Additional Testing Performed**
Tested Manually by uploading changes to private org - https://dev.azure.com/v-dmerugu0683/Sample%20Project/_build/results?buildId=1023&view=logs&j=4ef10708-f316-5e1d-20e0-afbca60c92cd&t=77245893-7256-55ca-6acf-ef97785d1a31
image

---

### **Documentation Changes Required** (Yes / No)  
yes - https://learn.microsoft.com/en-us/nuget/consume-packages/package-source-mapping

---
### **Related Issues**
https://github.com/microsoft/azure-pipelines-tasks/issues/21507
---

### **Checklist**
- [x] Related issue linked (if applicable)
- [x] Package version was bumped — see [versioning guide](https://semver.org/)
- [x] Verified the package behaves as expected
[...]

PR #635, opened 2026-06-26
Fix duplicate feed-* key in packageSourceMapping sync when a manual feed- entry already exists

### LINKED ISSUE #634: syncPackageSourceMappingKeys creates duplicate feed-* keys when a manual feed-* workaround entry already exists (regression from #543)
## Title

`syncPackageSourceMappingKeys` creates duplicate `feed-*` keys when a manual `feed-*` workaround entry already exists (regression from #543)

## Task / package

`packaging-common` — `nuget/NuGetConfigHelper2.ts`
Affects `DotNetCoreCLI@2` (restore with `feedsToUse: config`) and `NuGetCommand@2`.

## Summary

PR #543 (commit c85ac42a) added `syncPackageSourceMappingKeys`, which rewrites internal `<packageSource key="X">` entries in `<packageSourceMapping>` of the generated temp config to the prefixed form `feed-X`.

The rewrite does not check whether a `<packageSource key="feed-X">` entry **already exists** in `<packageSourceMapping>`. Many existing configs contain such a manual entry, because adding both `X` and `feed-X` to the mapping was the long-standing community workaround (NuGet/Home#11406) from before #543 existed.

When both are present, the temp config ends up with two identical `feed-X` keys and restore fails:

```
error : PackageSourceMapping is enabled and there are multiple package sources
associated with the same key(s): feed-<name>.
Path: /home/vsts/work/1/Nuget/tempNuGet_XXXXXX.config
```

This surfaced without any user config change, as the updated task rolled out to hosted agents.

## Minimal repro

`nuget.config` committed to the repo (internal Azure Artifacts feed + the historical manual `feed-` workaround entry):

```xml
<configuration>
  <packageSources>
    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
    <add key="MyFeed" value="https://pkgs.dev.azure.com/org/_packaging/MyFeed/nuget/v3/index.json" />
  </packageSources>
  <packageSourceMapping>
    <packageSource key="MyFeed">
      <package pattern="MyCompany.*" />
    </packageSource>
    <!-- manual workaround entry from NuGet/Home#11406 -->
    <packageSource key="feed-MyFeed">
      <package pattern="MyCompany.*" />
    </packageSource>
    <packageSource key="nuget.org">
      <package pattern="*" />
    </packageSource>
  </packageSourceMapping>
</configuration>
```

```yaml
- task: NuGetAuthenticate@1
- task: DotNetCoreCLI@2
  inputs:
    command: restore
    feedsToUse: config
    nugetConfigPath: nuget.config
```

**Expected:** restore succeeds; `MyFeed` is rewritten to `feed-MyFeed` exactly once.
**Actual:** temp config has two `feed-MyFeed` keys → restore fails with "multiple package sources associated with the same key(s)".

## Root cause

In `syncPackageSourceMappingKeys`, the rewrite branch renames `key` to `feed-<name>` unconditionally (`a.key = prefixedKey`) without verifying the prefixed key isn't already declared by another node. The existing `startsWith("feed-")` check only makes the function idempotent against its own output, not against pre-existing user-authored `feed-*` entries.

## Proposed fix

Before rewriting, collect the set of keys already present in `<packageSourceMapping>`. If the target `feed-<name>` key already exists, skip the rename and leave the original node untouched, instead of unconditionally overwriting `a.key`. This is the smallest possible change: one extra guard in the existing loop, no node removal, forward iteration preserved. A stale mapping entry whose key no longer matches a `<packageSources>` entry is ignored by NuGet, so leaving it in place is harmless. Diff in the linked PR.

andrew102:
### What

`syncPackageSourceMappingKeys` (added in #543) renamed internal `<packageSource key="X">`
mapping entries to `feed-X` without checking whether a `feed-X` entry already existed.
Configs that carried the historical manual `feed-X` workaround (NuGet/Home#11406) ended up
with two identical `feed-X` keys, breaking restore with:

> PackageSourceMapping is enabled and there are multiple package sources associated with
> the same key(s): feed-<name>.

### Change

- Build a set of keys already present in `<packageSourceMapping>` before rewriting.
- If the target `feed-<name>` key already exists, **skip the rename** and leave the
  original node untouched, instead of unconditionally overwriting `a.key` (which produced
  a duplicate `feed-<name>` key).
- Keep the existing idempotency guard for nodes already prefixed.

Nothing is added or removed — the only behavioural change is one extra guard that prevents
a rename when it would collide with an existing key. Forward iteration is preserved.

### Patch

`common-npm-packages/packaging-common/nuget/NuGetConfigHelper2.ts` — replaces the rewrite loop:

```typescript
    // Keys already present in <packageSourceMapping>. Used to detect when a
    // target prefixed key already exists (e.g. a user added a manual
    // "feed-<name>" entry as the historical workaround from
    // https://github.com/NuGet/Home/issues/11406). Renaming blindly in that
    // case produces two identical "feed-<name>" keys and NuGet fails with
    // "multiple package sources associated with the same key(s)".
    const existingKeys = new Set<string>();
    for (const src of psmSources) {
        const k = src?.$?.key;
        if (typeof k === "string") { existingKeys.add(k.trim()); }
    }

    let changed = false;
    for (const src of psmSources) {
        const a = src?.$;
        if (!a || typeof a.key !== "string") { continue; }
        const originalKey = a.key.trim();

        // Already prefixed -> leave as-is (idempotent).
        if (originalKey.startsWith("feed-")) { continue; }

        const prefixedKey = originalToPrefixed.get(originalKey);
        if (!prefixedKey || originalKey === prefixedKey) { continue; }

        // If the prefixed key is already declared, renaming would create a
        // duplicate key. Leave the original node untouched instead. NuGet
        // ignores a mapping entry whose key no longer matches a packageSources
        // entry, so the stale unprefixed node is harmless.
        if (existingKeys.has(prefixedKey)) {
            tl.debug(`PackageSourceMapping-Sync: '${prefixedKey}' already present; skipping rename of '${originalKey}'.`);
            continue;
        }

        tl.debug(`PackageSourceMapping-Sync: key '${originalKey}' -> '${prefixedKey}'`);
        a.key = prefixedKey;
        changed = true;
    }
```

### Notes

This deliberately leaves the original unprefixed node in place rather than removing it.
A mapping entry referencing a key that no longer exists in `<packageSources>` (because the
source was renamed to `feed-<name>`) is ignored by NuGet, not an error — so the stale node
is harmless, and skipping is the smallest, lowest-risk change. The correct match
(`feed-<name>` → pattern) is still provided by the pre-existing entry.

### Testing

Repro config with both `MyFeed` and `feed-MyFeed` in `<packageSourceMapping>` now restores
successfully; temp config contains a single `feed-MyFeed` key.

Fixes #634

### DIFF
--- a/common-npm-packages/packaging-common/nuget/NuGetConfigHelper2.ts
+++ b/common-npm-packages/packaging-common/nuget/NuGetConfigHelper2.ts
@@ -193,23 +193,42 @@ export class NuGetConfigHelper2 {
     const psmSources = psmNode?.packageSource;
     if (!Array.isArray(psmSources) || psmSources.length === 0) { tl.debug("no <packageSource> children."); return; }
 
+    // Keys already present in <packageSourceMapping>. Used to detect when a
+    // target prefixed key already exists (e.g. a user added a manual
+    // "feed-<name>" entry as the historical workaround from
+    // https://github.com/NuGet/Home/issues/11406). Renaming blindly in that
+    // case produces two identical "feed-<name>" keys and NuGet fails with
+    // "multiple package sources associated with the same key(s)".
+    const existingKeys = new Set<string>();
+    for (const src of psmSources) {
+        const k = src?.$?.key;
+        if (typeof k === "string") { existingKeys.add(k.trim()); }
+    }
+
     let changed = false;
     for (const src of psmSources) {
         const a = src?.$;
         if (!a || typeof a.key !== "string") { continue; }
         const originalKey = a.key.trim();
 
-        // If this mapping key maps to an internal feed now prefixed, rewrite it.
+        // Already prefixed -> leave as-is (idempotent).
+        if (originalKey.startsWith("feed-")) { continue; }
+
         const prefixedKey = originalToPrefixed.get(originalKey);
-        if (prefixedKey && originalKey !== prefixedKey) {
-            tl.debug(`PackageSourceMapping-Sync: key '${originalKey}' -> '${prefixedKey}'`);
-            a.key = prefixedKey;
-            changed = true;
+        if (!prefixedKey || originalKey === prefixedKey) { continue; }
+
+        // If the prefixed key is already declared, renaming would create a
+        // duplicate key. Leave the original node untouched instead. NuGet
+        // ignores a mapping entry whose key no longer matches a packageSources
+        // entry, so the stale unprefixed node is harmless.
+        if (existingKeys.has(prefixedKey)) {
+            tl.debug(`PackageSourceMapping-Sync: '${prefixedKey}' already present; skipping rename of '${originalKey}'.`);
             continue;
         }
 
-        // Already prefixed -> leave as-is (idempotent).
-        if (originalKey.startsWith("feed-")) { continue; }
+        tl.debug(`PackageSourceMapping-Sync: key '${originalKey}' -> '${prefixedKey}'`);
+        a.key = prefixedKey;
+        changed = true;
     }
 
     if (changed) {
--- a/common-npm-packages/packaging-common/package-lock.json
+++ b/common-npm-packages/packaging-common/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "azure-pipelines-tasks-packaging-common",
-    "version": "3.273.1",
+    "version": "3.273.2",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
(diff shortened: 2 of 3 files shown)


andrew102:

pt., 26 cze 2026 o 12:27 microsoft-github-policy-service[bot] <
***@***.***> napisał(a):

PR #637, opened 2026-07-14
azure-arm-rest: add gated Kudu log VSO-command sanitizer (MSRC 125166)

onetocny:
### **Description**
Fixes ICM 31000000654080 / MSRC 125166: `KuduServiceUtility` echoes remote Kudu/Oryx deployment and build log lines via `console.log` without sanitization. The Azure Pipelines agent scans all task stdout for `##vso[` and executes matching logging commands, so an attacker-controlled `##vso[...]` command embedded in application **source** (e.g. surfaced through an Oryx server-side build log via an `npm` postinstall script) executes on the agent at the pipeline's trust level — without the attacker ever needing pipeline access.

Adds `sanitizeKuduLogForConsole` (+ helpers `findVsoCommands`, `emitKuduLogSanitizerDetectionTelemetry`) to `azure-arm-app-service-kudu.ts`, directly after the `Kudu` class. Consuming tasks (AzureRmWebAppDeployment@3-5, AzureWebApp@1, AzureWebAppContainer@1, AzureFunctionApp@1/@2, AzureFunctionAppContainer@1) will be updated in a follow-up PR to reference this package (via a packed tgz) instead of their own local copy of the sanitizer.

Behavior is gated by the `getPipelineFeature('EnableKuduLogVsoCommandSanitization')` pipeline feature:
- **Feature disabled (default):** detection-only. Suspicious `##vso[...]` patterns are detected and reported via `telemetry.publish` (`event=KuduLogVsoCommandsDetected`, `enforced=false`) so real-world impact can be measured before enforcing.
- **Feature enabled:** matched `##vso[...]` patterns are neutralized before the line reaches `console.log`, and telemetry reports `enforced=true`.

E2E-validated in a live Azure DevOps pipeline against all 8 consuming task versions (uploaded as private builds): the sanitizer correctly detects and reports the injected `##vso` commands from server-side Oryx build logs in dry-run (telemetry-only) mode with the feature flag off, exactly as designed.

---

### **Package Name**
`azure-pipelines-tasks-azure-arm-rest`

---

**Low.** New exported functions only; no existing exports or `Kudu` class behavior changed. Default-disabled feature flag means no behavior change until the flag is explicitly enabled per-pipeline. Telemetry-only by default.

---

- [ ] Unit tests added or updated
- [x] Manual tests performed

_No package-level unit test suite exists for `azure-arm-rest` (no `test` script in `package.json`). Verified with `npm run build` (clean `tsc` compile) and end-to-end pipeline testing described above._

---

Packed this package as a `.tgz` and referenced it via `file:` in all 8 consuming tasks, rebuilt each task, uploaded private builds to a test Azure DevOps organization, and ran a purpose-built E2E repro pipeline that plants a malicious `##vso[...]` command in application source deployed through App Service Oryx builds. Confirmed via task logs that the new sanitizer correctly detects the injected commands and publishes `KuduLogVsoCommandsDetected` telemetry with `enforced=false` while the feature flag is off (default/dry-run state).

---

No.

---

### **Dependencies**
None added.

---

### **Checklist**
- [x] Related issue linked (if applicable) — ICM 31000000654080 / MSRC 125166
- [x] Package version was bumped — see [versioning guide](https://semver.org/) (3.277.0 → 3.277.1)

### DIFF
--- a/common-npm-packages/azure-arm-rest/azure-arm-app-service-kudu.ts
+++ b/common-npm-packages/azure-arm-rest/azure-arm-app-service-kudu.ts
@@ -842,14 +842,24 @@ export class Kudu {
 // If no ##vso[...] pattern is present in the text, this is a no-op (no telemetry emitted for
 // clean logs).
 
-const VSO_COMMAND_REGEX = /##vso\[([a-zA-Z0-9_.]+)/gi;
+// Used only to decide whether there is anything to do at all (telemetry and/or enforcement).
+// Intentionally NOT restricted to any command-name charset, so the kill switch below does not
+// depend on the agent's current command-name parsing rules - any "##vso[" (or leading "##[")
+// occurrence is treated as something to neutralize when enforcement is on.
+const VSO_COMMAND_PRESENCE_REGEX = /##vso\[/i;
+
+// Used only to name detected commands for telemetry. This charset is a best-effort label and
+// intentionally must NOT be used to decide whether enforcement/telemetry should run - see
+// VSO_COMMAND_PRESENCE_REGEX above for that gate.
+const VSO_COMMAND_NAME_REGEX = /##vso\[([a-zA-Z0-9_.]+)/gi;
 const KUDU_LOG_SANITIZER_TELEMETRY_AREA = 'TaskHub';
 const KUDU_LOG_SANITIZER_ENFORCE_PIPELINE_FEATURE = 'EnableKuduLogVsoCommandSanitization';
 
+// For telemetry naming only - see comment on VSO_COMMAND_NAME_REGEX above.
 function findVsoCommands(text: string): string[] {
     const found = new Set<string>();
     let match: RegExpExecArray | null;
-    const regex = new RegExp(VSO_COMMAND_REGEX);
+    const regex = new RegExp(VSO_COMMAND_NAME_REGEX);
     while ((match = regex.exec(text)) !== null) {
         found.add(match[1].toLowerCase());
     }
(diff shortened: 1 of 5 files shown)


common-npm-packages/azure-arm-rest/azure-arm-app-service-kudu.ts
```diff
@@ -801,3 +801,113 @@ export class Kudu {
...
+ *        'AzureFunctionApp') so telemetry can be attributed correctly.
+ *
+ * - Detection telemetry always runs: if `text` contains any ##vso[...] command sequence, it is
+ *   always reported via telemetry.publish, regardless of the pipeline feature state.
+ * - EnableKuduLogVsoCommandSanitization off (default today): text is returned completely
+ *   unmodified - telemetry-only, zero behavior change for customers.
+ * - EnableKuduLogVsoCommandSanitization on: enforce. Neutralizes the "##vso[" (and leading
+ *   "##[") trigger sequences so the agent can no longer parse them as logging commands.
+ */
+export function sanitizeKuduLogForConsole(text: string, telemetryFeature: string): string {
```
manolerazvan, line 903: Since this is the core security helper and it's a pure function, it would be really nice to add a small unit test suite for it. The package doesn't have a `test` script today, but it already pulls in the mocha and nock types, so adding one spec is cheap. A few cases worth locking down: clean text passes through untouched, `##vso[task.setvariable]` is detected, enforcement neutralizes `##vso[` when the feature is on, the text is left unchanged (but telemetry still fires) when the feature is off, and a crafted payload cannot break out of the telemetry command envelope. That keeps the regex and enforcement contract protected against future edits.

common-npm-packages/azure-arm-rest/azure-arm-app-service-kudu.ts
```diff
@@ -801,3 +801,113 @@ export class Kudu {
...
+    if (!text) {
+        return text;
+    }
+
+    const detectedCommands = findVsoCommands(text);
+    if (detectedCommands.length === 0) {
```
manolerazvan: One thing worth tightening: enforcement only runs when `findVsoCommands` returns a match, and that function is intentionally restricted to the `[a-zA-Z0-9_.]` command-name charset for telemetry. It happens to be safe today because the agent will not dispatch a command whose name falls outside that set, but it couples the kill switch to a telemetry-shaped regex and to the agent's current parsing behavior. Consider gating enforcement on the raw `##vso[` presence (or simply always running the replace when the feature is on), and keep the charset regex purely for naming the telemetry. That makes the neutralization self-evidently correct without depending on agent internals.

manolerazvan (approved):

common-npm-packages/azure-arm-rest/azure-arm-app-service-kudu.ts
onetocny: Addressed.

common-npm-packages/azure-arm-rest/azure-arm-app-service-kudu.ts
onetocny: Addressed.

manolerazvan (approved):

manolerazvan (approved):

PR #641, opened 2026-08-06
webdeployment-common: harden XDT config transformations and block xdt:Import assembly loading

nemanjarogic:
### **Description**

XDT (XML Document Transform) config transformations in this package are executed with the bundled `ctt.exe` (Config Transformation Tool, which embeds `Microsoft.Web.XmlTransform` **v1.6.0.51029**). The XDT engine supports the `<xdt:Import>` element, which loads a .NET assembly — **by name (`assembly=`) or by path (`path=`)** — and then constructs and executes custom transform/locator types from it (`System.Reflection.Assembly.Load` / `Assembly.LoadFrom`). Because transform files come from the deployment package/artifact (**customer/attacker‑influenceable content**), a crafted transform file can cause **arbitrary assembly loading and code execution on the pipeline agent** during deployment. A relative `path=` even resolves next to the transform file, so the assembly can ship inside the same package.

This code path is reached by the major deployment tasks that consume this package: **AzureRmWebAppDeployment V3/V4/V5, AzureWebApp / AzureFunctionApp and their Container variants, FileTransform V1/V2, and IISWebAppDeploymentOnMachineGroup**. They all funnel through this package’s single `ctt.exe` invocation (`applyXdtTransformation`, reached via `fileTransformationsUtility` → `basic`/`specialXdtTransformation`), so fixing it here closes the vector once for every consuming task.

**What this PR does** — it validates the transform file **before** it is handed to `ctt.exe`:

- **Blocks `<xdt:Import>`** — namespace‑aware, so it catches any namespace prefix, an aliased prefix, and a default (prefix‑less) XDT namespace.
- **Allows only the engine’s built‑in `xdt:Transform` / `xdt:Locator` types** — any custom (non‑built‑in) type is rejected. The allow‑list was verified by reflecting the concrete (non‑abstract) `Transform` / `Locator` subclasses out of the bundled `ctt.exe` binary (10 transforms + 4 locators for v1.6.0.51029), so it matches the engine exactly.
- **Fails closed on malformed / hostile XML** — the file is decoded via the package’s existing `detectFileEncoding()` utility (single source of truth with the XML/JSON variable‑substitution paths; UTF‑8 and UTF‑16 LE only) and parsed with the existing `@xmldom/xmldom` dependency whose error handler throws — so DTD/entity smuggling and undeclared entities are rejected rather than silently mis‑parsed.

**Secure‑by‑default, with an opt‑out escape hatch.** The block is **on by default**. Setting the pipeline variable **`AZP_ALLOW_UNSAFE_XDT_TRANSFORMS=true`** restores the previous (unsafe) behavior for authors who legitimately depend on custom XDT transforms. This is deliberately gated on a variable that **only the pipeline author can set** — the deployment‑package supplier cannot — and every bypass emits a build **warning** and a **telemetry** event. It lets us ship the hardening now and gives operators an immediate self‑service switch if a regression surfaces, without a hotfix.

**Telemetry.** The validator emits `##vso[telemetry.publish area=TaskHub;feature=XdtTransformationSecurity]` with `{ "result": "blocked" | "bypassed", "reason": "import" | "customTransform" | "customLocator" | "optOut" }` on both enforcement and opt‑out, so we can size real‑world usage of `xdt:Import` and watch the opt‑out rate during rollout.

Built‑in, XML‑only transforms are unaffected. No public API/signature changes.

**Related work item:** https://portal.microsofticm.com/imp/v5/incidents/details/31000000672943/summary

---

### **Package Name**

`azure-pipelines-tasks-webdeployment-common` (version `4.274.2` → `4.274.3`)

---

### **Risk Assessment**

**Medium.**

- **Backward compatibility for the common case is high.** Only the engine’s built‑in transforms/locators are allowed, and the allow‑list matches exactly the built‑in types in the bundled `ctt.exe` (verified by reflection). The engine resolves type names case‑sensitively, so anything the allow‑list rejects would already fail in `ctt.exe` today. The existing positive transform test is unchanged.
- **The deliberate behavior change is that `<xdt:Import>` (and custom transform/locator types) are now blocked** (secure‑by‑default). Pipelines that rely on custom XDT transforms via `xdt:Import` will fail with a clear, actionable error — a breaking change for that (expected small) population.
- **Blast radius is broad** — many high‑usage deployment tasks consume this package — which is why the change is scoped narrowly to pre‑execution validation, preserves all built‑in behavior, and ships with an author‑controlled opt‑out plus telemetry so impact can be measured and mitigated ring‑by‑ring.

---

- [x] Manual tests performed

---

- **Unit tests** in `Tests/L1XdtTransform.ts` — now 10 XDT cases:
  - Rejections: `xdt:Import` by `path`, `xdt:Import` by `assembly` name, `xdt:Import` with an alternate namespace prefix, custom `xdt:Transform` type, custom `xdt:Locator` type.
  - **New:** DTD/entity smuggling of `xdt:Import` **fails closed**; `xdt:Import` declared in the **default XDT namespace** is blocked; **built‑in transforms/locators with arguments** (`Match(key)` / `SetAttributes(value)`) are allowed; the **opt‑out** (`AZP_ALLOW_UNSAFE_XDT_TRANSFORMS=true`) restores legacy behavior.
  - The original positive test (a real XDT transformation through `ctt.exe`) is unchanged and still applies built‑ins (`SetAttributes` / `Insert` / `Remove`).
- **Vulnerability reproduced** against the bundled `ctt.exe`: `xdt:Import` triggers assembly load and executes a custom transform; validation blocks it before `ctt.exe` runs.
- **Allow‑list verified** by reflecting the concrete `Transform` / `Locator` subclasses embedded in `ctt.exe` (v1.6.0.51029); confirmed the engine is case‑sensitive (no case‑related regression) and that built‑in transforms still apply.
- **Validator (`@xmldom/xmldom`) hardening checks:** DTD/entity‑based smuggling and undeclared entities fail closed (rejected); default‑namespace and alternate‑prefix `xdt:Import` are detected.
- **Opt‑out + telemetry verified** on the compiled module: `true` (case/space‑tolerant) bypasses validation and restores legacy behavior; anything else (incl. `false`, `1`, `yes`, empty) blocks; telemetry events fire with the expected `result`/`reason` on both the block and the bypass paths.
- **Encoding end‑to‑end:** UTF‑8 (incl. BOM) and UTF‑16 LE decode/validate correctly; unsupported encodings (UTF‑16 BE / UTF‑32) are rejected with a clear message.

---

### **Documentation Changes Required**

**Yes.** Customer‑facing guidance / release notes should note that (will create a new PR):
1. `xdt:Import` (loading custom transform assemblies) and non‑built‑in `xdt:Transform`/`xdt:Locator` types are no longer supported by default in XDT config transformations, with the recommended alternative (pre‑apply custom transforms outside the task, or use built‑in transforms); and
2. the `AZP_ALLOW_UNSAFE_XDT_TRANSFORMS` variable exists as an explicit, **unsafe** opt‑out and what risk it re‑introduces.

Target the "File transforms and variable substitution reference" docs and the affected task pages. No public API/signature changes.

---

### **Dependencies**

No new runtime dependencies. 

---

### **Checklist**

- [x] Package version was bumped — `4.274.2` → `4.274.3` (see [versioning guide](https://semver.org/))


common-npm-packages/webdeployment-common/xdttransformationutility.ts
```diff
@@ -38,6 +67,165 @@ export function applyXdtTransformation(sourceFile: string, transformFile: string
...
+
+function validateXdtTransformFile(transformFile: string): void {
+    if (isUnsafeXdtTransformAllowed()) {
+        // Opt-out escape hatch: restores the pre-hardening behavior for pipeline authors who
+        // legitimately depend on custom XDT transforms.
+        if (!unsafeXdtTransformWarningEmitted) {
+            tl.warning(tl.loc('XdtTransformationSecurityValidationDisabled', transformFile));
+            unsafeXdtTransformWarningEmitted = true;
+        }
+        publishXdtSecurityTelemetry('bypassed', 'optOut');
```
ivanduplenskikh: nit: Should we deduplicate telemetry here when we go through [the array of configs](https://github.com/microsoft/azure-pipelines-tasks-common-packages/blob/0f2423e1eed7733a4bd9d2209eabff7fd6acc6d3/common-npm-packages/webdeployment-common/xdttransformationutility.ts#L54)

nemanjarogic: Done.

ivanduplenskikh (approved):

PR #657, opened 2026-09-03
Validate Azure Container Registry host before registry authentication (opt-in)

rm-dmiri:
### **Description**

Adds an opt-in check that validates the Azure Container Registry (ACR) login server against the known ACR host suffixes before the registry authentication request in `docker-common`.

- New `registryhostvalidation.ts`: `isAllowedAcrHost()` accepts standard ACR login servers across public, sovereign, and air-gapped clouds (`.azurecr.io`, `.azurecr.us`, `.azurecr.cn`, `.azurecr.de`, `.azurecr.eaglex.ic.gov`, `.azurecr.microsoft.scloud`) and rejects inputs that are not a plain ACR hostname. `shouldBlockRegistryHost()` applies the feature gate.
- `ACRAuthenticationTokenProvider` runs the check before the registry authentication request.
- Adds the `InvalidRegistryHost` localized message (en-US only).

Gated by the pipeline feature **`AcrRegistryHostValidation`** (read via `tl.getPipelineFeature`), **default off**. Enable per pipeline with the variable `DistributedTask.Tasks.AcrRegistryHostValidation: true`. Behavior is unchanged when the feature is not enabled.

It is intentionally opt-in and kept as a permanent switch because the ACR login server is environment-dependent: some environments (for example Azure Stack Hub) use custom registry host suffixes that a fixed list cannot enumerate, so enabling validation there is left to the pipeline owner.

---

### **Package Name**

`azure-pipelines-tasks-docker-common`

---

**Low.** The new behavior is gated behind a feature that is off by default, so there is no change to existing behavior unless a pipeline explicitly enables it. The validation module is self-contained and covered by unit tests.

---

- [ ] Manual tests performed

---

Ran the `docker-common` L0 suite locally; all tests pass, including the new host-validation cases (accepted/rejected hosts across public, sovereign, and air-gapped clouds) and the feature-gate cases (on/off).

---

No.

---

### **Dependencies**

None. No new runtime dependencies are introduced.

---

### **Checklist**
- [ ] Related issue linked (if applicable)
- [x] Package version was bumped to 2.280.0 — see [versioning guide](https://semver.org/)

### DIFF
--- a/common-npm-packages/docker-common/Tests/acrProviderHostValidationTests.ts
+++ b/common-npm-packages/docker-common/Tests/acrProviderHostValidationTests.ts
@@ -1,0 +1,70 @@
+// Provider-level checks that the host gate is enforced at BOTH public entry points
+// (getToken and getAuthenticationToken). The agent temp/working-dir env must be set
+// before importing the provider, which transitively loads azure-arm-rest.
+process.env["SYSTEM_DEFAULTWORKINGDIRECTORY"] = process.env["SYSTEM_DEFAULTWORKINGDIRECTORY"] || require("os").tmpdir();
+process.env["AGENT_TEMPDIRECTORY"] = process.env["AGENT_TEMPDIRECTORY"] || require("os").tmpdir();
+
+import assert = require("assert");
+import * as tl from "azure-pipelines-task-lib/task";
+import ACRAuthenticationTokenProvider from "../registryauthenticationprovider/acrauthenticationtokenprovider";
+
+export function runAcrProviderHostValidationTests() {
+
+    const featureVariable = "DistributedTask.Tasks.AcrRegistryHostValidation";
+
+    function setFeature(enabled: boolean): void {
+        tl.setVariable(featureVariable, enabled ? "true" : "false");
+    }
+
+    // registerNameValue is a raw host string (not JSON), so registryURL === host.
+    function providerFor(host: string): any {
+        return new ACRAuthenticationTokenProvider("endpoint", host);
+    }
+
+    afterEach(() => {
+        setFeature(false);
+    });
+
+    describe("getAuthenticationToken() (service-principal path)", () => {
+        it("feature ON + non-ACR host: throws", () => {
+            setFeature(true);
+            assert.throws(() => providerFor("other.example.com").getAuthenticationToken());
+        });
+
+        it("feature ON + valid ACR host: does not throw", () => {
+            setFeature(true);
+            assert.doesNotThrow(() => providerFor("contoso.azurecr.io").getAuthenticationToken());
+        });
+
+        it("feature OFF + non-ACR host: does not throw (behavior unchanged)", () => {
+            setFeature(false);
+            assert.doesNotThrow(() => providerFor("other.example.com").getAuthenticationToken());
+        });
+
+        it("feature ON + empty host: does not throw (returns null)", () => {
+            setFeature(true);
+            assert.strictEqual(providerFor("").getAuthenticationToken(), null);
+        });
+    });
+
+    describe("getToken() (gate runs before the auth-scheme dispatch, so it covers all schemes)", () => {
+        it("feature ON + non-ACR host: rejects before dispatching on scheme", (done) => {
+            setFeature(true);
+            providerFor("other.example.com").getToken().then(
+                () => done(new Error("expected the call to be rejected")),
+                (err: any) => {
+                    try { assert.ok(err && String(err.message).length > 0); done(); }
+                    catch (e) { done(e); }
+                }
+            );
+        });
+
+        it("feature OFF + non-ACR host: does not reject at the gate (behavior unchanged)", (done) => {
+            setFeature(false);
+            providerFor("other.example.com").getToken().then(
+                () => done(),
+                (err: any) => done(err)
+            );
+        });
+    });
+}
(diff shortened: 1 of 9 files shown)


common-npm-packages/docker-common/registryauthenticationprovider/registryhostvalidation.ts
```diff
@@ -0,0 +1,63 @@
...
+    );
+}
+
+export const AcrHostValidationFeatureName = "AcrRegistryHostValidation";
+
+export function guardRegistryHost(registryURL: string, endpointId: string, scheme: string): void {
+    if (!tl.getPipelineFeature(AcrHostValidationFeatureName)) {
+        return;
+    }
+    if (!registryURL || isAllowedAcrHost(registryURL)) {
```
ivanduplenskikh: Should we simplify the check here since we're checking **registryURL** in **isAllowedAcrHost**?

common-npm-packages/docker-common/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
     "name": "azure-pipelines-tasks-docker-common",
-    "version": "2.279.0",
+    "version": "2.280.0",
```
ivanduplenskikh: We should update both: package.json and package-lock.json

common-npm-packages/docker-common/registryauthenticationprovider/registryhostvalidation.ts
rm-dmiri: You are right. I can move that to validation function. Thanks!

common-npm-packages/docker-common/Tests/acrRegistryHostValidationTests.ts
```diff
@@ -0,0 +1,138 @@
...
+            return out;
+        }
+
+        afterEach(() => setEnforce(false));
+
+        it("enforcing + non-ACR host: warns with the host, endpoint id, and scheme", () => {
+            setEnforce(true);
+            const out = capture(() => guardRegistryHost("other.example.com", "endpoint-abc", "ServicePrincipal"));
+            assert.ok(out.indexOf("task.issue type=warning") !== -1, "expected a warning");
+            assert.ok(out.indexOf("other.example.com") !== -1, "warning should name the host");
```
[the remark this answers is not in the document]
rm-dmiri: Reworked these assertions to compare against the localized message built via tl.loc(...), a computed value, and to assert the function result directly, instead of checking for a hostname substring - so the incomplete-substring pattern no longer applies. Should clear on the next scan.

common-npm-packages/docker-common/registryauthenticationprovider/registryhostvalidation.ts
rm-dmiri: Good point - simplified as suggested. Dropped the separate empty-value check and let isAllowedAcrHost be the single decision (isAllowedAcrHost("") is false), so an empty/absent login server now flows into the same invalid path and is logged instead of being silently skipped (an empty ACR login server is not a legitimate value anyway). Done in the latest push.

common-npm-packages/docker-common/package-lock.json
rm-dmiri: Correction: landed this as 2.280.2 (patch within the current sprint 280, above base 2.280.1) rather than 2.281.0, and regenerated package-lock.json so both version fields match. Also merged main to clear the version-line conflict.

common-npm-packages/docker-common/registryauthenticationprovider/registryhostvalidation.ts
```diff
@@ -0,0 +1,63 @@
+"use strict";
+
+import * as tl from "azure-pipelines-task-lib/task";
+
```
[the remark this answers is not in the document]
rm-dmiri: Fixed - added tl.setResourcePath(path.join(__dirname, "..", "module.json"), true) at module load so the localized strings resolve even when this module is imported on its own.

common-npm-packages/docker-common/Tests/acrProviderHostValidationTests.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    // Raw (non-JSON) host string, so registryURL === host.
+    function providerFor(host: string): any {
+        return new ACRAuthenticationTokenProvider("endpoint", host);
+    }
+
+    afterEach(() => {
+        setFeature(false);
+    });
```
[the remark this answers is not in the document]
rm-dmiri: The provider reads those endpoint parameters with optional = true, so unmocked they return undefined rather than throwing - that is why the allowed-host cases resolve without a live endpoint. The tests exercise the host-validation branch, and the full L0 suite is green (64 tests).

common-npm-packages/docker-common/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "azure-pipelines-tasks-docker-common",
-    "version": "2.280.0",
+    "version": "2.281.0",
```
starkmsu: It should be 2.280.1 I think

rm-dmiri: Oh, this is wrong, let me fix it. Thanks for noticing

starkmsu (approved):

PR #659, opened 2026-09-07
azure-arm-rest: make Kudu log sanitizer whitelist-aware (MSRC 125166)

onetocny:
### **Description**

Changes the Kudu deployment-log sanitizer from neutralizing **all** `##vso[` logging commands to neutralizing only **non-whitelisted** ones, mirroring the AzureFileCopy approach in [azure-pipelines-tasks#22451](https://github.com/microsoft/azure-pipelines-tasks/pull/22451).

`sanitizeKuduLogForConsole` (in `common-npm-packages/azure-arm-rest/azure-arm-app-service-kudu.ts`) is used by the App Service / Function deployment tasks to print remote Kudu/Oryx build logs. Previously, when enforcement was on it escaped every `##vso[` command, blocking legitimate logging commands too. It now consults the server-delivered allow-list.

Related ADO Work Item: [AB#2440075](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2440075) · MSRC 125166 / ICM 31000000654080 (CWE-77 command injection).

Behavior when the `EnableKuduLogVsoCommandSanitization` pipeline feature is **on**:
- Allow-list read from `AGENT_ALLOWEDLOGGINGCOMMANDS` (pipeline variable `agent.allowedLoggingCommands`), comma-separated, case-insensitive.
- **Empty/unset allow-list → allow all** (escape nothing) — consistent with #22451.
- Non-empty allow-list → escape `##vso[<cmd>` → `##_vso[<cmd>` only for commands **not** on the list; whitelisted commands pass through verbatim.
- **Leading `##[` bracket (formatting) sequences are never modified** — only `##vso[` commands are ever neutralized.
- Feature **off** → telemetry-only, text unmodified (no behavior change).

Telemetry: the emitted payload carries `enforced`, `allowlistPresent`, and `escapedCount` so the rollout signal can distinguish "actively protecting" (`escapedCount > 0`) from an allow-all no-op (`enforced` but `escapedCount == 0`). Detection events (`KuduLogVsoCommandsDetected` / `KuduLogVsoCommandsSanitized`) are unchanged.

No public API change (`sanitizeKuduLogForConsole(text, telemetryFeature)`).

---

### **Package Name**
`azure-pipelines-tasks-azure-arm-rest` (bumped `3.279.5` → `3.280.0`)

---

**Low.** Change is confined to the enforcement branch of a single function that is only active when the `EnableKuduLogVsoCommandSanitization` feature is enabled (still not broadly enabled). With the allow-all-on-empty default, absence of an allow-list preserves today's behavior of not modifying output. No signature change, so consumers only need a dependency bump to adopt.

---

- [x] Manual tests performed

---

Extended `Tests/L0-kudu-log-sanitizer-tests.ts` with whitelist cases: whitelisted command preserved, non-whitelisted escaped, case-insensitive match, empty-list-allows-all, telemetry `allowlistPresent`/`escapedCount` assertions, and leading `##[` sequences left untouched (both with and without an allow-list). Ran the sanitizer suite: **10 passing** (isolated via `mocha --grep`). Other azure-arm-rest L0 timeouts observed are pre-existing flakes (AKS / OpenSSL checks) unrelated to this change.

---

**No.**

---

### **Dependencies**
None introduced. Consuming tasks (AzureRmWebAppDeploymentV3/V4/V5, AzureWebAppV1, AzureWebAppContainerV1, AzureFunctionAppV1/V2, AzureFunctionAppContainerV1) will pick up the new behavior via a dependency bump to this package once published — no task source changes required.

---

### **Checklist**
- [x] Package version was bumped — see [versioning guide](https://semver.org/)
- [x] CLA signed (if applicable) — see [Contributor License Agreement](https://cla.opensource.microsoft.com)

### DIFF
--- a/common-npm-packages/azure-arm-rest/azure-arm-app-service-kudu.ts
+++ b/common-npm-packages/azure-arm-rest/azure-arm-app-service-kudu.ts
@@ -855,6 +855,42 @@ const VSO_COMMAND_NAME_REGEX = /##vso\[([a-zA-Z0-9_.]+)/gi;
 const KUDU_LOG_SANITIZER_TELEMETRY_AREA = 'TaskHub';
 const KUDU_LOG_SANITIZER_ENFORCE_PIPELINE_FEATURE = 'EnableKuduLogVsoCommandSanitization';
 
+// Whitelist-aware neutralization (aligns with azure-pipelines-tasks PR #22451 for AzureFileCopy).
+// The set of logging commands that must NOT be neutralized is delivered by the server as the
+// read-only pipeline variable 'agent.allowedLoggingCommands', surfaced on the agent as the
+// environment variable AGENT_ALLOWEDLOGGINGCOMMANDS (comma-separated, case-insensitive).
+//   - whitelist empty/unset   -> escape nothing; every "##vso[" command is allowed through
+//                                (consistent with AzureFileCopy #22451 feature-off semantics).
+//   - command NOT whitelisted -> its "##vso[" prefix is escaped to "##_vso[" so the agent does
+//                                not execute it (the line is still printed as readable text).
+//   - command whitelisted     -> left unchanged so legitimate usage keeps working.
+// This whitelist only ever narrows the set of "##vso[" commands that get escaped; it is consulted
+// only when enforcement (EnableKuduLogVsoCommandSanitization) is on. Leading "##[" sequences carry
+// no command name to match against the whitelist and are always neutralized when enforcing.
+const KUDU_LOG_SANITIZER_ALLOWED_COMMANDS_ENV_VAR = 'AGENT_ALLOWEDLOGGINGCOMMANDS';
+
+// Captures the command name after "##vso[" so it can be tested against the whitelist. The charset
+// matches VSO_COMMAND_NAME_REGEX; a "##vso[" with no following name yields an empty capture which
+// is never whitelisted (and is therefore escaped when a whitelist is present).
+const VSO_COMMAND_REPLACE_REGEX = /##vso\[([a-zA-Z0-9_.]*)/gi;
+
+// Parses AGENT_ALLOWEDLOGGINGCOMMANDS into a case-insensitive (lower-cased) set of command names.
+// Returns an empty set when the variable is unset/blank, which callers treat as "allow all".
+function getAllowedLoggingCommands(): Set<string> {
+    const allowed = new Set<string>();
+    const raw = process.env[KUDU_LOG_SANITIZER_ALLOWED_COMMANDS_ENV_VAR];
+    if (!raw || !raw.trim()) {
+        return allowed;
+    }
+    for (const command of raw.split(',')) {
+        const trimmed = command.trim().toLowerCase();
+        if (trimmed) {
+            allowed.add(trimmed);
+        }
+    }
+    return allowed;
+}
+
 // For telemetry naming only - see comment on VSO_COMMAND_NAME_REGEX above.
 function findVsoCommands(text: string): string[] {
     const found = new Set<string>();
@@ -897,8 +933,10 @@ function emitKuduLogSanitizerDetectionTelemetry(feature: string, enforced: boole
  *   always reported via telemetry.publish, regardless of the pipeline feature state.
  * - EnableKuduLogVsoCommandSanitization off (default today): text is returned completely
  *   unmodified - telemetry-only, zero behavior change for customers.
- * - EnableKuduLogVsoCommandSanitization on: enforce. Neutralizes the "##vso[" (and leading
- *   "##[") trigger sequences so the agent can no longer parse them as logging commands.
+ * - EnableKuduLogVsoCommandSanitization on: enforce, but whitelist-aware. Any leading "##["
+ *   sequence is neutralized. "##vso[" commands are neutralized only when they are NOT on the
+ *   allow-list delivered via AGENT_ALLOWEDLOGGINGCOMMANDS (agent.allowedLoggingCommands); when the
+ *   allow-list is empty/unset, no "##vso[" command is neutralized (allow all).
  */
 export function sanitizeKuduLogForConsole(text: string, telemetryFeature: string): string {
     if (!text || !VSO_COMMAND_PRESENCE_REGEX.test(text)) {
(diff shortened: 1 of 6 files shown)


wawanawna:
**`azure-arm-app-service-kudu.ts` - `sanitizeKuduLogForConsole` (empty allow-list / allow-listed pass-through):**

The whitelist reads `AGENT_ALLOWEDLOGGINGCOMMANDS`, but that variable is unset by default - and unset means the agent honors *every* logging command. So with the feature ON and no list configured (today's default), this escapes nothing for `##vso[` (only the leading `##[`). When a list is present, allow-listed commands are passed through and the agent executes them.

For Kudu/Oryx logs the content is attacker-influenceable (app source, `npm` postinstall, transitive deps), and there's no legitimate reason to honor `##vso[` from it. Compared to the previous "escape everything" behavior this is a step down: in the default state nothing is neutralized, and any allow-listed command can still be injected and run.

Can we confirm with the security owner that (1) the rollout populates a restrictive `agent.allowedLoggingCommands` before enforcement is enabled, and (2) "enforcement on + empty list = no `##vso[` neutralization" is acceptable for this source? If not, keeping the blanket escape on the Kudu path is the safer default.

wawanawna:
[minor] **`azure-arm-app-service-kudu.ts` - detection telemetry (`enforced` field):**

`enforced:true` is emitted whenever the feature is on - even when the allow-list is empty and nothing actually gets escaped (the new "allows all ... even though the feature is on" test asserts exactly that). Since this telemetry drives the rollout decision, it can't distinguish "actively protecting" from "allow-all no-op".

Could we add something like `escaped:<count>` or `allowlistPresent:true/false` so the signal reflects whether protection is really active?

onetocny:
> **`azure-arm-app-service-kudu.ts` - `sanitizeKuduLogForConsole` (empty allow-list / allow-listed pass-through):**
> 
> The whitelist reads `AGENT_ALLOWEDLOGGINGCOMMANDS`, but that variable is unset by default - and unset means the agent honors _every_ logging command. So with the feature ON and no list configured (today's default), this escapes nothing for `##vso[` (only the leading `##[`). When a list is present, allow-listed commands are passed through and the agent executes them.
> 
> For Kudu/Oryx logs the content is attacker-influenceable (app source, `npm` postinstall, transitive deps), and there's no legitimate reason to honor `##vso[` from it. Compared to the previous "escape everything" behavior this is a step down: in the default state nothing is neutralized, and any allow-listed command can still be injected and run.
> 
> Can we confirm with the security owner that (1) the rollout populates a restrictive `agent.allowedLoggingCommands` before enforcement is enabled, and (2) "enforcement on + empty list = no `##vso[` neutralization" is acceptable for this source? If not, keeping the blanket escape on the Kudu path is the safer default.

In this case the remote service is App Service. I think that has a higher level of trust than custom virtual machines. I have discussed the topic with Eric and point is that we want to treat all these use cases in a same way. We have similar fixes in AzureFileCopy and PowershellOnTargetMachine tasks. Pipeline tasks team is working on similar fix for Ssh task. The point is that all of these should use the whitelist from single config framework value. Based on the telemetry we already have for AzureRmWebAppDeployment task there is small number of usages (mainly task.setvariable) which was identified by Eric and Subham as legitimate cases.

I kept the enforcement FF for purpose - it may happen that allowedLoggingCommands will be already initialized at the time when new version of AzureRmWebAppDeployment task will be deployed (I want to start the roll out today to fix the AzureFileCopy vulnerability) and I want to have some back door in case something wont work as expected. When allowedLoggingCommands is empty we treat it as the fix is not enabled. This is how it works in AFC and I wanted to keep it in sync even here.

onetocny:
@wawanawna good call — the telemetry can now tell active protection apart from an allow-all no-op.

Pushed in 682fcf4. Enforcement now runs first and counts neutralizations, and the detection payload gained two fields:

- `allowlistPresent` — whether a non-empty `agent.allowedLoggingCommands` allow-list was configured.
- `escapedCount` — how many sequences were actually neutralized on this call.

So `enforced:true` + `escapedCount:0` + `allowlistPresent:false` is unambiguously the allow-all no-op you flagged, while `escapedCount > 0` means protection was actually applied — this also correctly reflects the always-neutralized leading `##[` case even when the allow-list is empty. Extended the L0 tests to assert both fields (empty-allow-list → `escapedCount:0`/`allowlistPresent:false`; leading-bracket → `escapedCount:1`); full `azure-arm-rest` suite is green (38 passing). Updated the PR description accordingly.

onetocny:
> [minor] **`azure-arm-app-service-kudu.ts` - detection telemetry (`enforced` field):**
> 
> `enforced:true` is emitted whenever the feature is on - even when the allow-list is empty and nothing actually gets escaped (the new "allows all ... even though the feature is on" test asserts exactly that). Since this telemetry drives the rollout decision, it can't distinguish "actively protecting" from "allow-all no-op".
> 
> Could we add something like `escaped:<count>` or `allowlistPresent:true/false` so the signal reflects whether protection is really active?

@wawanawna good call — the telemetry can now tell active protection apart from an allow-all no-op.

Pushed in 682fcf4. Enforcement now runs first and counts neutralizations, and the detection payload gained two fields:

- `allowlistPresent` — whether a non-empty `agent.allowedLoggingCommands` allow-list was configured.
- `escapedCount` — how many sequences were actually neutralized on this call.

So `enforced:true` + `escapedCount:0` + `allowlistPresent:false` is unambiguously the allow-all no-op you flagged, while `escapedCount > 0` means protection was actually applied — this also correctly reflects the always-neutralized leading `##[` case even when the allow-list is empty. Extended the L0 tests to assert both fields (empty-allow-list → `escapedCount:0`/`allowlistPresent:false`; leading-bracket → `escapedCount:1`); full `azure-arm-rest` suite is green (38 passing). Updated the PR description accordingly.

common-npm-packages/azure-arm-rest/azure-arm-app-service-kudu.ts
```diff
@@ -897,8 +945,13 @@ function emitKuduLogSanitizerDetectionTelemetry(feature: string, enforced: boole
...
+ * - EnableKuduLogVsoCommandSanitization on: enforce, but whitelist-aware. Any leading "##["
+ *   sequence is neutralized. "##vso[" commands are neutralized only when they are NOT on the
+ *   allow-list delivered via AGENT_ALLOWEDLOGGINGCOMMANDS (agent.allowedLoggingCommands); when the
+ *   allow-list is empty/unset, no "##vso[" command is neutralized (allow all).
+ * - The emitted telemetry carries `enforced`, `allowlistPresent` and `escapedCount` so the rollout
+ *   signal can tell "actively protecting" (escapedCount > 0) apart from an allow-all no-op
+ *   (enforced but escapedCount == 0).
  */
 export function sanitizeKuduLogForConsole(text: string, telemetryFeature: string): string {
     if (!text || !VSO_COMMAND_PRESENCE_REGEX.test(text)) {
```
starkmsu, line 957: The early return only checks for `##vso[`, so input containing only a leading `##[` sequence skips enforcement entirely and is returned unchanged even when the feature is enabled. The current test masks this because its payload includes both forms. Could we include `^##\[` (multiline) in the presence check and add a test whose payload contains only the leading-bracket sequence?

onetocny: I have decided not to escape ##[  at all as those are harmless (see latest commit). This is also in line with how AzureFileCopy behaves. See https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#formatting-commands.

starkmsu (approved):

starkmsu (approved):

PR #665, opened 2026-09-16
Harden XDT validation against encoding differentials

nagarajku:
### **Description**
This PR hardens XDT transformation validation against encoding differences between the Node.js validator and the bundled .NET `ctt.exe` transformation tool.

The validator now:

- Validates the declared encoding of both source and transform XML files.
- Blocks encodings that can cause the validator and transformation tool to interpret XML differently.
- Verifies UTF-16 encoding consistency.
- Requires namespace-aware XDT nodes.
- Fails closed when encoding validation cannot be completed.
- Reads and validates the exact transform bytes passed to the transformation process.

This prevents unsafe XDT content from bypassing validation before transformation.

### **Package Name**
azure-pipelines-tasks-webdeployment-common

---

### **Risk Assessment** (High)  
This is a security hardening change affecting XML transformations used by deployment tasks. The main behavior change is that unsupported or inconsistent XML encodings are rejected instead of being processed. Existing UTF-8, UTF-16LE, ASCII-compatible, and supported code-page scenarios remain supported.

---

- [X ] Unit tests added or updated
- [ X] Manual tests performed

Added regression coverage for:

- Unsafe encoding declarations.
- Source/transform encoding mismatches.
- UTF-16 consistency.
- Empty and malformed XML declarations.
- Multiple encoding variants.
- Namespace-aware XDT validation.
- Existing supported encoding scenarios.

---

- Package build completed successfully.
- Full test suite passed: **52 tests**.
- Manually validated behavior against the bundled `ctt.exe`.
- Confirmed supported UTF-8 transformations continue to work.
- Confirmed unsafe encoding scenarios are blocked before transformation.

---

### **Documentation Changes Required** (No)  

Security-sensitive behavior is documented in code comments and user-facing localization strings.
---

### **Dependencies**
No new dependencies were introduced. The package version was updated to 4.280.4

---

### **Checklist**
- [ ] Related issue linked (if applicable)
- [X ] Package version was bumped — see [versioning guide](https://semver.org/)
- [X ] Verified the package behaves as expected

---


common-npm-packages/webdeployment-common/xdttransformationutility.ts
```diff
@@ -84,11 +162,170 @@ function validateXdtTransformFile(transformFile: string): void {
         return;
     }
 
-    const xmlContent = readTransformFile(transformFile);
-    const transformDocument = parseTransformFile(transformFile, xmlContent);
+    // ctt.exe decodes the transform file using the encoding declared by the *source* document, so the
```
nemanjarogic: The comment here (and the MSRC 139444 test) states that ctt.exe decodes the transform using the source document's declared encoding. That doesn't match the XDT engine:` XmlTransformation(transformFile)` loads the transform through `XmlFileInfoDocument.Load` → `StreamReader(file, true)` + `XmlTextReader`, so the transform is decoded from its own BOM + its own `<?xml encoding>`. The source's encoding is only used for `Save()` (output).

Can you confirm the exact ctt.exe wrapper behavior with a concrete repro (i.e. which file's declaration actually flips `+ADw`-→`<)`?

If it's the transform's declaration, then `validateXdtEncodingDeclaration(sourceFile)` is not closing a real vector and only adds a functional regression for sources that declare non-ASCII-transparent encodings and the MSRC test (source=utf-7, transform=no declaration) isn't actually exploitable and should be rewritten to put the utf-7 declaration on the transform.

common-npm-packages/webdeployment-common/xdttransformationutility.ts
```diff
@@ -28,6 +28,84 @@ const builtInXdtLocatorTypes = [
...
+//
+// Deliberately excluded: utf-7 and the ISO-2022 / HZ-GB-2312 family (escape sequences synthesise
+// ASCII characters), every EBCDIC code page (0x00-0x7F map to entirely different characters), and
+// utf-32 (not modelled by detectFileEncoding). UTF-16 is alignment-sensitive rather than
+// ASCII-transparent, so the UTF-16 aliases are additionally subject to the explicit decoder
+// agreement check in validateUtf16DecoderAgreement. UTF-16 BE is excluded outright because
+// detectFileEncoding always rejects it, so this validator can never model a UTF-16 BE transform.
+const asciiTransparentEncodings = buildAsciiTransparentEncodingSet();
+
+function buildAsciiTransparentEncodingSet(): { [encodingName: string]: boolean } {
```
nemanjarogic, line 80: The soundness of the whole gate depends on every entry in `asciiTransparentEncodings` being genuinely ASCII-transparent (0x00–0x7F → ASCII, and no multi-byte trail/escape byte can synthesize or swallow a delimiter < 0x40).

This is the security anchor and it's a broad list. Please add a focused test that, for each multi-byte CJK entry (shift_jis/gbk/gb18030/big5/euc-*), confirms .NET's decoder cannot produce a `<`/`>`/`&`/`=` from any non-ASCII byte sequence, mirroring the reflection-based verification used for the transform/locator type allowlist so a future edit to this list can't quietly reintroduce a differential.

common-npm-packages/webdeployment-common/xdttransformationutility.ts
nagarajku: Thanks for the review, pushed fixes to address this:
 **Source-file encoding check was invalid. ** Confirmed via the dotnet/xdt source that ctt.exe loads the transform file completely independently of the source document, so checking the source's declared encoding didn't close any real vector and only risked rejecting valid files. Removed it, corrected the comments, and rewrote the affected tests to target the transform file (the actual attack surface).

common-npm-packages/webdeployment-common/xdttransformationutility.ts
nagarajku: Thanks for the review, pushed fixes to address this: 
**CJK allowlist now has empirical backing.** Added a verification harness that uses .NET's own Encoding tables to confirm shift_jis/gbk/big5/euc-jp/euc-kr can't synthesize an XML delimiter from non-ASCII bytes, checked in the results, and added a regression test that guards the allowlist against future silent additions.

nemanjarogic (approved):

